Issue #10, opened 2016-01-16
Accessing FIXSession inside FIXStatusListener


jheusser:
In the example Initiator a `FIXStatusListener` is passed to the `FIXSession` as anonymous class (see [here](https://github.com/jvirtanen/philadelphia/blob/master/philadelphia-initiator/src/main/java/org/jvirtanen/philadelphia/initiator/Initiator.java#L33)). For more complicated `FIXStatusListeners` (or for multiple active `FIXSessions`) it's not convenient to pass it as anonymous class and thus I ended up with the repetitive and bit awkward pattern

```Java
BaseFixStatusListener statusListener = new BaseFixStatusListener();
FIXSession fixSession = new FIXSession(channel, fixConfig, inbound, statusListener);
statusListener.setFIXSession(fixSession);
```

Where the `BaseFixStatusListener` has a `setFIXSession` method. If you always want to access the `FIXSession` in a `FixStatusListener` it might make sense to add such a setter method to the interface, or similar.

jvirtanen:
Yes, this is something I have been thinking about.

I think the cleanest approach would be to augment `FIXMessageListener` and `FIXStatusListener` methods to contain a `session` argument. An example:

```
void message(FIXSession session, FIXMessage message)
```

What do you think? @penberg, any thoughs?

penberg:
@jvirtanen Sounds reasonable to me.

jheusser:
@jvirtanen sounds good yes, also allows to potentially share a `FIXStatusListener` between multiple `FIXSessions`. I think the status listener has more immediate need for the session than the message listener, but probably makes sense for consistency to add the argument to both.

jvirtanen:
@jheusser Good point again. I agree that we should do it only for `FIXStatusListener`, at least for now. Would you be interested in submitting a pull request?

jheusser:
@jvirtanen Ok, will submit a PR adding a `FIXSession` parameter to each method in `FIXStatusListener`.

jvirtanen:
Great, thanks! :+1:

jheusser:
@jvirtanen Doing this now. Maybe it is also required in `FIXMessageListener` as the `MessageHandler` inside `FIXSession` is also passing down events to the session `FIXStatusListener`. Or I pass the session in via constructor in the `MessageHandler`. You decide what's best.

jvirtanen:
[...]

PR #13, opened 2016-01-21
Accessing FIXSession inside FIXStatusListener

jheusser:
As discussed in #10 

Ran tests and initiators/acceptors for ping pong test.

### DIFF
--- a/philadelphia/src/main/java/org/jvirtanen/philadelphia/FIXStatusListener.java
+++ b/philadelphia/src/main/java/org/jvirtanen/philadelphia/FIXStatusListener.java
@@ -10,57 +10,64 @@ public interface FIXStatusListener {
     /**
      * Receive an indication to close the connection.
      *
+     * @param session the Session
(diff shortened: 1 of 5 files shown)


philadelphia/src/main/java/org/jvirtanen/philadelphia/FIXStatusListener.java
```diff
@@ -10,57 +10,64 @@
     /**
      * Receive an indication to close the connection.
      *
+     * @param session the Session
```
jvirtanen: A small nit, but for uniformity, could you use lowercase for "session" (cf. "message" below)? Thanks for bearing with me. :grin:

jvirtanen:
Looks very good!

jheusser:
Makes sense, updated with javadoc adjustment.

jvirtanen:
Merged, thank you!

jheusser:
:+1:

PR #39, opened 2017-03-28
Reduce size of FIXSession.MessageHandler method

jheusser:
This is an issue / PR, the commits just make it clearer.

Every FIX message passes through the `FIXSession.MessageHandler.message` method, so it's on the critical path of most FIX applications. Unfortunately with 391 bytes in size it's the only method in Philadelphia larger than the 325 bytes default jvm hot method inline size.

The commit below reduces the size of the method to 315 bytes which allows inlining. Running the Philadelphia ping pong test before:
```
@ 275   com.paritytrading.philadelphia.FIXSession$MessageHandler::message (391 bytes)   hot method too big
```
after:
```
@ 275   com.paritytrading.philadelphia.FIXSession$MessageHandler::message (315 bytes)   inline (hot)
```

The removal of the check at line 553 should be fine as any message with 2 bytes message type is not the most common message type and caught by the default-case in the switch statement (assuming no such message type would match 0-A in the first byte which would be retarded). As we are not doing the length check anymore we directly access the char at offset 0, and if there's no match it passes it downstream.

### DIFF
--- a/philadelphia/src/main/java/com/paritytrading/philadelphia/FIXSession.java
+++ b/philadelphia/src/main/java/com/paritytrading/philadelphia/FIXSession.java
@@ -538,24 +538,13 @@ public class FIXSession implements Closeable {
             }
 
             if (msgSeqNum < rxMsgSeqNum) {
-                if (msgType.length() != 1 || msgType.asChar() != SequenceReset) {
-                    FIXValue possDupFlag = message.findField(PossDupFlag);
-
-                    if (possDupFlag == null || possDupFlag.asChar() != 'Y')
-                        statusListener.tooLowMsgSeqNum(FIXSession.this, msgSeqNum, rxMsgSeqNum);
-                }
-
+                handleTooLowMsgSeqNum(message, msgSeqNum, msgType);
                 return;
             }
 
             rxMsgSeqNum++;
 
-            if (msgType.length() != 1) {
-                downstream.message(message);
-                return;
-            }
-
-            switch (msgType.asChar()) {
+            switch ((char) msgType.byteAt(0)) {
             case Heartbeat:
                 handleHeartbeat();
                 break;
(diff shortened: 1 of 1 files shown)


jheusser:
Just when I sent this off I remembered [msgType=AD](http://www.onixs.biz/fix-dictionary/4.4/msgType_AD_6568.html) .... something to think about.

jvirtanen:
Good work, @jheusser. I think that the change in the message type handling can indeed be problematic due to the overlap in message type values. However, the sequence reset handling could be optimized a bit by moving more of [the logic](https://github.com/paritytrading/philadelphia/blob/master/philadelphia/src/main/java/com/paritytrading/philadelphia/FIXSession.java#L527) into the `handleSequenceReset` method.

jheusser:
Yes, agree. Factoring out the gap fill logic and putting the message type handling back in makes sense. 

If you like I can do that. I could move the gap fill logic into `handleSequenceReset` and remove the boolean parameter?

jvirtanen:
Yes, that would be great. The boolean parameter is indeed no longer needed, but I think `handleSequenceReset` will then have to return a boolean indicating whether or not the sequence reset was handled (for the first call site).

jheusser:
Method is now at 314 bytes, with `handleSequenceReset` refactored and message type check back in. Let me know how this looks.

philadelphia/src/main/java/com/paritytrading/philadelphia/FIXSession.java
```diff
@@ -524,12 +524,8 @@ public void message(FIXMessage message) throws IOException {
             }
 
             if (msgType.length() == 1 && msgType.asChar() == SequenceReset) {
-                FIXValue gapFillFlag = message.findField(GapFillFlag);
-
-                if (gapFillFlag == null || gapFillFlag.asChar() != 'Y') {
-                    handleSequenceReset(message, true);
+                if(handleSequenceReset(message))
```
jvirtanen: A small nit: Could you retain the space after `if`:
```java
if (handleSequenceReset(message))
```

philadelphia/src/main/java/com/paritytrading/philadelphia/FIXSession.java
```diff
@@ -555,7 +545,7 @@ public void message(FIXMessage message) throws IOException {
                 return;
             }
 
-            switch (msgType.asChar()) {
+            switch (msgType.byteAt(0)) {
```
jvirtanen: :+1:

philadelphia/src/main/java/com/paritytrading/philadelphia/FIXSession.java
```diff
@@ -583,6 +573,15 @@ public void message(FIXMessage message) throws IOException {
             }
         }
 
+        private void handleTooLowMsgSeqNum(FIXMessage message, long msgSeqNum, FIXValue msgType) throws IOException {
```
jvirtanen: Another small nit: Here I would prefer `msgType` before `msgSeqNum` in the argument list as that is their order in the FIX message header.

jvirtanen:
Thanks, @jheusser!

PR #48, opened 2018-03-05
philadelphia-core: Remove public modifier on interface

jheusser:


jvirtanen:
Thank you @jheusser, a nice clean-up! ✨

PR #49, opened 2018-03-05
philadelphia-core: Use System.arraycopy to set FIXValue

jheusser:
- Use `System.arraycopy` to set value
- Add method to `perf-test` as was not included in tests yet

### DIFF
--- a/tests/perf-test/src/main/java/com/paritytrading/philadelphia/perf/FIXValueBenchmark.java
+++ b/tests/perf-test/src/main/java/com/paritytrading/philadelphia/perf/FIXValueBenchmark.java
@@ -57,22 +57,22 @@ public class FIXValueBenchmark {
 
         timestamp = new MutableDateTime(2015, 9, 24, 9, 30, 5, 250);
 
-        charValue = new FIXValue(32);
+        charValue = new FIXValue(64);
         charValue.setChar('Y');
 
-        intValue = new FIXValue(32);
+        intValue = new FIXValue(64);
         intValue.setInt(123);
 
-        floatValue = new FIXValue(32);
+        floatValue = new FIXValue(64);
         floatValue.setFloat(12.30, 2);
 
-        stringValue = new FIXValue(32);
+        stringValue = new FIXValue(64);
         stringValue.setString("FOO");
 
-        dateValue = new FIXValue(32);
+        dateValue = new FIXValue(64);
         dateValue.setDate(date);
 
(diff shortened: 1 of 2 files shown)


tests/perf-test/src/main/java/com/paritytrading/philadelphia/perf/FIXValueBenchmark.java
```diff
@@ -45,6 +45,8 @@
 
     private FIXValue timestampValue;
 
+    private FIXValue timestampValueDestination;
```
jvirtanen: Maybe this could be simply called e.g. `copyValue`?

tests/perf-test/src/main/java/com/paritytrading/philadelphia/perf/FIXValueBenchmark.java
```diff
@@ -155,4 +159,9 @@ public void setTimestamp() {
         timestampValue.setTimestamp(timestamp, true);
     }
 
+    @Benchmark
+    public void setTimestampByValue() {
+        timestampValueDestination.set(timestampValue);
+    }
```
jvirtanen: Maybe this test could be simply called e.g. `setToValue`?

tests/perf-test/src/main/java/com/paritytrading/philadelphia/perf/FIXValueBenchmark.java
```diff
@@ -73,8 +75,10 @@ public void prepare() {
         timeOnlyValue = new FIXValue(32);
         timeOnlyValue.setTimeOnly(timeOnly, true);
 
-        timestampValue = new FIXValue(32);
+        timestampValue = new FIXValue(64);
```
jvirtanen: Why the increase in the value capacity?

jheusser: The default `FIXValue` is actually 64 via `FIXConfig`, and as it's a perf test I thought to set it to the default.

jvirtanen: Good point. :+1: Maybe we could the same for all `FIXValue` instances in this benchmark? :wink:

jvirtanen:
Great stuff! Thank you, @jheusser! 🙌

PR #52, opened 2018-04-26
philadelphia-client: Add 'wait' command

jheusser:
Adds a 'wait' command which waits for a specific MsgType. Usage is:
```
wait 8
```

which waits for a message with `35=8` .

This is useful where the 'sleep' command is not enough to coordinate interactions with an acceptor.


jvirtanen:
Great, this looks like a great addition! 🙌 Adding some comments inline. :mag:

applications/client/src/main/java/com/paritytrading/philadelphia/client/command/WaitCommand.java
```diff
@@ -0,0 +1,56 @@
+package com.paritytrading.philadelphia.client.command;
+
+import com.paritytrading.philadelphia.client.TerminalClient;
+import com.paritytrading.philadelphia.client.message.Message;
+import com.paritytrading.philadelphia.client.message.Messages;
+import org.eclipse.collections.api.list.ImmutableList;
+
+import java.util.Scanner;
```
jvirtanen: In Philadelphia, the imports are usually grouped together in alphabetical order. 🤓

applications/client/src/main/java/com/paritytrading/philadelphia/client/command/WaitCommand.java
```diff
@@ -0,0 +1,56 @@
...
+
+class WaitCommand implements Command {
+    private static final long WAIT_TIME_MILLIS = 500;
```
jvirtanen: What if we set this to 100 or 50 milliseconds to speed up script execution? 🤔

applications/client/src/main/java/com/paritytrading/philadelphia/client/command/WaitCommand.java
```diff
@@ -0,0 +1,56 @@
...
+            try {
+                Thread.sleep(WAIT_TIME_MILLIS);
+
+                if(lastMsgType(client.getMessages()).equals(msgType))
+                    break;
+
+            } catch(InterruptedException e) {
+                break;
+            }
+        }
```
jvirtanen: Perhaps we could maintain the existing spacing between a control word (`while`, `if`, `catch`) and opening parenthesis. 🤓

applications/client/src/main/java/com/paritytrading/philadelphia/client/command/WaitCommand.java
```diff
@@ -0,0 +1,56 @@
...
+    @Override
+    public String getUsage() {
+        return "wait <MsgType>";
+    }
+
+    private String lastMsgType(Messages messages) {
+        ImmutableList<Message> collect = messages.collect();
+
+        return collect.get(collect.size() - 1).getMsgType();
+    }
```
jvirtanen: This will fail if we haven't yet received any messages. What about:
```java
if (messages.isEmpty())
    return null;

return messages.getLast().getMsgType();
```

With the above, we could then simply switch the check in the loop:
```java
if (msgType.equals(getLastMsgType(client.getMessages())))
    break;
```

Finally, perhaps this could be called `getLastMsgType` to fit in with the existing methods. :wink:

applications/client/src/main/java/com/paritytrading/philadelphia/client/command/WaitCommand.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    public String getName() {
+        return "wait";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Wait for a specific MsgType";
```
jvirtanen: In all documentation, we tend to use a form like "MsgType(35)" for FIX tags. Perhaps the description and usage of this command could use the same form?
- Wait for a specific MsgType(35)
- `wait <MsgType(35)>`

applications/client/src/main/java/com/paritytrading/philadelphia/client/command/WaitCommand.java
jheusser: Yeah I have no affinity to that syntax so I never see that. Adding 👍

jheusser:
Points addressed.

Issue #53, opened 2018-05-04
Expose constructor in FIXSession


jheusser:
It would be nice if the constructor would be public in `FIXSession` to pass in external clock:
```
FIXSession(Clock clock, SocketChannel channel, FIXConfig config, FIXMessageListener listener, FIXStatusListener statusListener)
```

jvirtanen:
Yes, good idea. :+1:

jheusser:
Also related to that, the class `SystemClock` is probably superfluous and can be replaced with `System::currentTimeMillis` usage these days.

jvirtanen:
Yes, definitely. That is a remnant from the Java 7 times. 🙂

PR #54, opened 2018-05-06
philadelphia-core: Make FIXSession constructor public

jheusser:
- Make `FIXSession` constructor public
- Remove superfluous `SystemClock` class

Closes https://github.com/paritytrading/philadelphia/issues/53


libraries/core/src/main/java/com/paritytrading/philadelphia/FIXSession.java
```diff
@@ -71,7 +71,7 @@
 
     private StringBuilder currentTimestamp;
 
-    FIXSession(Clock clock, SocketChannel channel, FIXConfig config, FIXMessageListener listener,
+    public FIXSession(Clock clock, SocketChannel channel, FIXConfig config, FIXMessageListener listener,
```
jvirtanen: Perhaps this could have a Javadoc comment like the existing public constructor.

Also, the `Clock` interface should be made public and amended with Javadoc comments. For example:
```java
/**
 * A clock.
 */
public interface Clock {

    /**
     * Get the current time in milliseconds.
     *
     * @return the current time in milliseconds
     * @see System#currentTimeMillis
     */
    long currentTimeMillis();

}
```

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXSession.java
```diff
@@ -138,7 +138,7 @@
      */
     public FIXSession(SocketChannel channel, FIXConfig config, FIXMessageListener listener,
             FIXStatusListener statusListener) {
-        this(SystemClock.INSTANCE, channel, config, listener, statusListener);
+        this(System::currentTimeMillis, channel, config, listener, statusListener);
```
jvirtanen: Nice clean-up! :sparkles:

jvirtanen:
Great work, @jheusser! 🙌

PR #55, opened 2018-05-08
philadelphia-core: Add FIXMessage#findIndex method

jheusser:
Like `FIXMessage#findField` but returns the index of the given tag.

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java
@@ -89,6 +89,22 @@ public class FIXMessage {
     }
 
     /**
+     * Get the index of the first instance of a field with the specified tag.
+     *
+     * @param tag the tag
+     * @return the index or -1 if there are no instances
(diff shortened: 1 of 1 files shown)


libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java
```diff
@@ -88,6 +88,23 @@ public FIXValue findField(int tag) {
...
 
+    /**
+     * Get the index of the value container of the first instance of a field with the
+     * specified tag.
+     *
+     * @param tag the tag
+     * @return the index or -1 if there are no instances
+     *   of a field with the specified tag
+     */
+    public int findIndex(int tag) {
```
jvirtanen: Maybe just: "Get the index of the first instance of a field with the specified tag." What do you think? 🤔

jheusser: Yes bit clumsy otherwise. 👍

jvirtanen:
Good addition, @jheusser! 🎉

PR #62, opened 2018-06-28
philadelphia-core: Remove 'FIXField' class

jvirtanen:
The `FIXMessage` class contains an array of `FIXField` objects, and the `FIXField` class contains a tag and a `FIXValue` object. Remove the `FIXField` class and make the `FIXMessage` class contain two arrays: one for tags and another for `FIXValue` objects.

Not only does this improve spatial locality within a `FIXMessage` object, it also makes working with the class simpler.


jvirtanen:
@jheusser Thoughts on the API changes?

jheusser:
Great change, tested this locally with an application that's using Philadelphia 👍 If anything code looks less verbose with new API.

jvirtanen:
I ran the performance tests on the weekend, they look pretty similar to `master`. After these changes, there might be some optimization opportunities in the methods of the `FIXMessage` class, but those are something to be investigated as further improvements.

PR #65, opened 2018-07-06
philadelphia-core: Add 'FIXMessage#fromString'

jvirtanen:


libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java
```diff
@@ -191,6 +192,34 @@ public void put(ByteBuffer buffer) {
         }
     }
 
+    /**
+     * Construct a new message container from a string representation.
+     *
+     * @param s a string representation
+     * @return a message container
+     */
+    public static FIXMessage fromString(String s) {
```
jheusser: I think looks good and is definitely useful. Only downside is that it's a bit an atypical method to most other methods/classes in the library (as it's allocating and only meant for test usage). Similarly to the `toString` method but it provides a method with `StringBuilder` at least. Could consider moving this to a helper method but no strong opinions.

jheusser: Agree, makes sense.

PR #69, opened 2018-08-13
philadelphia-core: Rename 'FIXSession' to 'FIXConnection'

jvirtanen:
- Rename `FIXSession` to `FIXConnection`
- Rename `FIXStatusListener` to `FIXConnectionStatusListener`

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/package-info.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/package-info.java
@@ -5,15 +5,15 @@
  * primary functions:</p>
  * <ul>
  * <li>data reception
- *   ({@link com.paritytrading.philadelphia.FIXSession#receive})</li>
+ *   ({@link com.paritytrading.philadelphia.FIXConnection#receive})</li>
  * <li>data transmission</li>
- * <li>session keep-alive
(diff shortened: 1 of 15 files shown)


libraries/core/src/main/java/com/paritytrading/philadelphia/package-info.java
```diff
@@ -5,15 +5,15 @@
...
  * <li>session keep-alive
- *   ({@link com.paritytrading.philadelphia.FIXSession#keepAlive})</li>
+ *   ({@link com.paritytrading.philadelphia.FIXConnection#keepAlive})</li>
  * </ul>
  *
- * <p>Data reception can run on one thread and data transmission and session
- * keep-alive on another without locking. Data transmission and session
- * keep-alive can run on different threads but require locking.</p>
+ * <p>Data reception can run on one thread and data transmission and
+ * session keep-alive on another without locking. Data transmission and
```
jheusser: Is the keep-alive per session or connection?

jvirtanen: A good point. For clarity, I will rename this to "connection keep-alive" as well. 👍

jvirtanen: Fixed.

PR #76, opened 2018-08-27
philadelphia-core: Simplify message header and trailer handling

jvirtanen:
- Use preformatted byte arrays for the BeginString(8), BodyLength(9), and CheckSum(10) tags
- When sending a message, instead of formatting the entire message header, only format the BodyLength(9) value


jheusser:
Great, this makes a lot sense. What does the perf tests say?

jvirtanen:
To stay on topic, I extracted the checksum optimizations outside the scope of this pull request. There is some further work to be done on that front.

In the performance tests, "background radiation" of the test setup currently clouds any conclusions to be made. There is certainly some further work to be done on _that_ front as well. 🙂

PR #77, opened 2018-09-17
philadelphia-core: Optimize 'FIXConnection#update' and 'FIXConnection#updateCompID'

jvirtanen:
Remove explicit `null` checks. This reduces the bytecode sizes from 45 bytes to 27 bytes, which is under HotSpot's default maximum inline size of 35 bytes.

**Note.** This is an API change as these methods will no longer throw an `InvalidStateException` but might throw a `NullPointerException` instead. Both being unchecked exceptions, I would argue that this is a backwards compatible API change. Thoughts?


jheusser:
Think this is a sensible change. If these tags aren't available then something else is more wrong than throwing a `NullPointerException` here.

PR #79, opened 2018-09-23
philadelphia-core: Optimize 'FIXTimestamps.append'

jvirtanen:
Using a buffer is roughly 16% faster than working with the `StringBuilder` directly.

~**Note.** This is an API change as this method will no longer be thread-safe. However, I would argue that this does not require bumping the major version number. Thoughts?~

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXTimestamps.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXTimestamps.java
@@ -7,6 +7,8 @@ import org.joda.time.ReadableDateTime;
  */
 public class FIXTimestamps {
 
+    private static final ThreadLocal<char[]> BUFFER = ThreadLocal.withInitial(() -> new char[21]);
+
     private FIXTimestamps() {
     }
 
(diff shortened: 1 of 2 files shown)


jheusser:
The change itself looks good, but shouldn't `FIXTimestamps` be non-static now, otherwise I believe you could not have multiple independent `FIXConnection` instances (in different threads)? So each `FIXConnection` instantiates a `FIXTimestamps` instance.

jvirtanen:
@jheusser That is a good point. I somehow managed to not think about multiple connections on separate threads at all. 😅 💦

Unfortunately we cannot change `FIXTimestamps.append` from a static method to a non-static method as that would be definitely a breaking public API change. What we _can_ do for the time being is to maintain thread-local `buffer` instances, please see the latest changes.

I think in the long run further gains should be possible by providing an alternative method that eliminates the `StringBuilder` altogether and works directly on e.g. a byte array or a `FIXValue`.

In Philadelphia 2.0.0, we should aim to remove these utility classes (`FIXTags`, `FIXTimestamps`, ...) from the public API altogether.

PR #81, opened 2018-09-29
philadelphia-core: Simplify keep-alive mechanism

jvirtanen:
After sending a TestRequest(1) message, consider the procedure successful on receiving any data from the other party, not just a Heartbeat(0) message.

The purpose of this change is to separate the keep-alive mechanism from the handling of administrative messages, making it easier to provide a way for applications to implement their own handling of administrative messages.


jvirtanen:
This is because the code is moved from the default administrative message handler to `FIXConnection`, and therefore custom administrative message handlers (an upcoming feature) will not need to contain any logic related to the keep-alive mechanism.

jvirtanen:
The Javadoc documentation for `FIXConnection#keepAlive` should be updated to reflect this change.

jheusser:
Ah, I get it now 👍

jvirtanen:
Now also the Javadoc documentation has been updated.

PR #83, opened 2018-10-15
philadelphia-core: Make 'FIXMessageParser' public

### LINKED ISSUE #51: make FIXMessageParser a public class
Is there any reason why a user shouldn't be able to use the FIXMessageParser without having to start an entire FIXSession? in case a users ByteBuffer isn't purely FIX messages but has FIX messages within the byte buffer to be parsed

jvirtanen:
Closes #51.


jvirtanen:
@jheusser Any thoughts on the proposed interface changes are very welcome.

PR #84, opened 2018-10-21
philadelphia-core: Add 'FIXValue#asBoolean' and 'FIXValue#setBoolean'

jvirtanen:


jvirtanen:
The boolean data type is the only one that lacks native `FIXValue` support.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
```diff
@@ -85,6 +85,32 @@ public void set(FIXValue value) {
...
     }
 
+    /**
+     * Get the value as a boolean.
+     *
+     * @return the value as a boolean
+     * @throws FIXValueFormatException if the value is not a boolean
+     */
+    public boolean asBoolean() {
+        if (length != 1)
```
jheusser: Is this a sufficient check that it's not a boolean? `notBoolean` and `notChar` are less meaningful than the other sanity checks (`notDigit` etc).

jheusser: Is this a sufficient check that it's not a boolean? `notBoolean` and `notChar` are less meaningful than the other sanity checks (`notDigit` etc).

jheusser: Is this a sufficient check that it's not a boolean? `notBoolean` and `notChar` are less meaningful than the other sanity checks (`notDigit` etc). Just thinking how much such sanity checks add.

jheusser: Sorry github was failing 😆.

jvirtanen: Good point on the general utility of these checks. 👍 

This one is modelled based on the `asChar` implementation. I added it, because we generally do check the length of the value. However, it would be good to think how much value these checks add altogether.

jvirtanen:
I still managed to squeeze in a small optimization. 🙂

PR #87, opened 2018-11-17
philadelphia-core: Clean up public interface

jvirtanen:
- Remove `FIXCheckSums` from public interface
- Remove `FIXTags` from public interface
- Remove `FIXTimestamps` from public interface

In addition to being used by the library internals, these utility classes are currently part of the public interface. Clean up the public interface and enable further optimizations by making these classes internal.

**Note.** This is a breaking change and scheduled for Philadelphia 2.0.0.

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXCheckSums.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXCheckSums.java
@@ -1,26 +1,9 @@
 package com.paritytrading.philadelphia;
 
-import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
 
-/**
- * Utilities for working with checksums.
- */
-public class FIXCheckSums {
+class FIXCheckSums {
 
-    private FIXCheckSums() {
-    }
-
-    /**
-     * Return the sum of bytes in a buffer.
-     *
-     * @param buffer a buffer
-     * @param offset the offset
-     * @param length the length
-     * @return the sum of bytes in the buffer
-     * @throws BufferUnderflowException if the length from the offset exceeds
-     *   the limit of the buffer
-     */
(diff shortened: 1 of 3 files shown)


libraries/core/src/main/java/com/paritytrading/philadelphia/FIXCheckSums.java
```diff
@@ -3,24 +3,8 @@
 import java.nio.BufferUnderflowException;
```
jheusser: Not used

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXCheckSums.java
```diff
@@ -3,24 +3,8 @@
...
-     * Return the sum of bytes in a buffer.
-     *
-     * @param buffer a buffer
-     * @param offset the offset
-     * @param length the length
-     * @return the sum of bytes in the buffer
-     * @throws BufferUnderflowException if the length from the offset exceeds
-     *   the limit of the buffer
-     */
     public static long sum(ByteBuffer buffer, int offset, int length) {
```
jheusser: Could be package private?

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXCheckSums.java
jvirtanen: Good catch! 👍 Fixed.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXCheckSums.java
jvirtanen: Yes, indeed. There are a few other places that have superfluous `public` modifiers within package-private classes. I thought about going through and cleaning these up in one go.

PR #98, opened 2019-02-06
philadelphia-perf-test: Add FIXMessageParser benchmark

khsergvl:


tests/perf-test/src/main/java/com/paritytrading/philadelphia/FIXMessageParserBenchmark.java
```diff
@@ -0,0 +1,65 @@
...
+
+        buffer.flip();
+
+        FIXConfig fixConfig = new FIXConfig.Builder().build();
+
+        messageParser = new FIXMessageParser(fixConfig, listener -> {});
+    }
+
+    private void format(FIXMessage message) {
+        message.addField(BeginString);
```
jvirtanen: Let's set the the field value as it would be for FIX 4.2:
```java
message.addField(BeginString).setString("FIX.4.2");
```

jvirtanen:
Thank you, @khsergvl, great work! 🙌 My apologies for the delay, I didn't get a notification that you had pushed changes since the previous review. 🙂

PR #99, opened 2019-03-19
philadelphia-core: Replace modulo with power of two bitmask

jheusser:
Minor performance improvement for checksum calculations by replacing power of two modulo operations with bitwise AND.

JIT might or might not do this optimisation eventually but order entry is frequently not entirely hot so this should be a clear and simple improvement.

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessageParser.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessageParser.java
@@ -126,7 +126,7 @@ public class FIXMessageParser {
                     continue;
 
                 // Garbled message
-                if (FIXCheckSums.sum(buffer, beginning, position - beginning + length) % 256
+                if ((FIXCheckSums.sum(buffer, beginning, position - beginning + length) & 0xff)
                         != checkSum.asCheckSum())
                     continue;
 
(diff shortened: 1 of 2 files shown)


libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessageParser.java
```diff
@@ -126,7 +126,7 @@ public boolean parse(ByteBuffer buffer) throws IOException {
                     continue;
 
                 // Garbled message
-                if (FIXCheckSums.sum(buffer, beginning, position - beginning + length) % 256
+                if ((FIXCheckSums.sum(buffer, beginning, position - beginning + length) & (256 - 1))
```
jvirtanen: Perhaps we could use `& 0xff` here, as the hexadecimal literals pair well with bitwise operations?

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
```diff
@@ -442,7 +442,7 @@ public long asCheckSum() {
      * @param c a checksum
      */
     public void setCheckSum(long c) {
-        setDigits(c % 256, 0, 3);
+        setDigits(c & (256 - 1), 0, 3);
```
jvirtanen: See above.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessageParser.java
jheusser: Good idea! 💥 Was too focussed on keeping the existing `256` visible with first `255` then `256 - 1`.

PR #101, opened 2019-04-24

jvirtanen:
Having the incoming CheckSum(10) check in a separate method makes message parsing 9.6% faster in the performance test when the incoming CheckSum(10) check is disabled and 3.7% faster when it is enabled.

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessageParser.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessageParser.java
@@ -124,29 +124,8 @@ public class FIXMessageParser {
 
             position = buffer.position();
 
-            if (checkSumEnabled) {
-                buffer.position(position + length);
-
-                // Garbled message
-                tag = FIXTags.get(buffer);
-                if (tag == 0)
-                    continue;
-
-                // Garbled message
-                if (!checkSum.get(buffer))
-                    continue;
-
-                // Garbled message
-                if (tag != CheckSum)
-                    continue;
-
-                // Garbled message
-                if ((FIXCheckSums.sum(buffer, beginning, position - beginning + length) & 0xff)
-                        != checkSum.asInt())
-                    continue;
-
-                buffer.position(position);
-            }
+            if (checkSumEnabled && !acceptCheckSum(buffer, beginning, position, length))
+                continue;
 
             int limit = buffer.limit();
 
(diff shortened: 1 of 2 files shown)


libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessageParser.java
```diff
@@ -124,29 +124,8 @@ public boolean parse(ByteBuffer buffer) throws IOException {
...
-                    continue;
-
-                // Garbled message
-                if ((FIXCheckSums.sum(buffer, beginning, position - beginning + length) & 0xff)
-                        != checkSum.asCheckSum())
-                    continue;
-
-                buffer.position(position);
-            }
+            if (checkSumEnabled && !checkSum(buffer, beginning, position, length))
```
jheusser: That's a good idea!

jheusser:
Will this be part of Philadelphia 1.1.0?

jvirtanen:
> Will this be part of Philadelphia 1.1.0?

No, that window is unfortunately now closed. 🙂 However, the aim is to get a more focused Philadelphia 1.2.0 not too long after this release.

jvirtanen:
Rebased on `master`.

jvirtanen:
@jheusser This is now ready for review. 🔍

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessageParser.java
```diff
@@ -124,29 +124,8 @@ public boolean parse(ByteBuffer buffer) throws IOException {
...
-                    continue;
-
-                // Garbled message
-                if ((FIXCheckSums.sum(buffer, beginning, position - beginning + length) & 0xff)
-                        != checkSum.asInt())
-                    continue;
-
-                buffer.position(position);
-            }
+            if (checkSumEnabled && !acceptCheckSum(buffer, beginning, position, length))
```
jheusser: is it "accept" or maybe "validate"? 🤷‍♂

jvirtanen: I think "accept" reads better when a method returns a Boolean value and "validate" when the acceptance or rejection is signaled some other way. 🙂

jheusser: Correct, true wordsmith!

PR #102, opened 2019-05-16
Fix coordinated omission

arturo-salinas:


jvirtanen:
Thank you, @arturo-salinas! I will have a look. 🔍

jvirtanen:
It looks like there is an issue in the way the ClOrdID(11) field is used to convey the send time:
```
philadelphia (bug-coordinated-omission) λ java -jar examples/initiator/philadelphia-initiator.jar localhost 4000 50000 10000
Warming up...
Benchmarking...
Results (n = 50000)

   50.00%: 181144540676.10 µs
   90.00%: 181144540676.10 µs
   99.00%: 181144540676.10 µs
   99.90%: 181144540676.10 µs
   99.99%: 181144540676.10 µs
  100.00%: 181144540676.10 µs
```

What about just putting the nanosecond timestamp directly to the ClOrdID(11) field to avoid the arithmetic on the receiving side?

arturo-salinas:
Yes, the arithmetic was wrong because the warming and benchmark phases were mixed up. One solution is to reset the index after the warming phase but your suggestion is much better. I implemented. Thank you!
```
java -jar philadelphia-initiator.jar localhost 4000 50000 10000
                                                   
Warming up...
Benchmarking...
Results (n = 50000)

   50.00%:     361.22 µs
   90.00%:  127139.84 µs
   99.00%:  170917.89 µs
   99.90%:  178388.99 µs
   99.99%:  179568.64 µs
  100.00%:  179961.86 µs

```

examples/initiator/src/main/java/com/paritytrading/philadelphia/initiator/Initiator.java
```diff
@@ -105,22 +114,28 @@ FIXConnection getTransport() {
...
 
-    void receive() throws IOException {
-        while (!received) {
-            if (connection.receive() < 0)
-                return;
-        }
+    void sendAndReceive(FIXMessage message, int orders) throws IOException {
+        for (long sentAtNanoTime = System.nanoTime(); receiveCount < orders; connection.receive()) {
+            if (System.nanoTime() >= sentAtNanoTime) {
+                FIXValue clOrdId = message.valueOf(ClOrdID);
```
jvirtanen: You could pass `FIXValue clOrdId` as a parameter to `sendAndReceive` to avoid invoking `FIXMessage#valueOf(int)` for each message: 🤓 
```java
void sendAndReceive(FIXMessage message, FIXValue clOrdId, int orders) throws IOException {
    ...
```

jvirtanen:
> Yes, the arithmetic was wrong because the warming and benchmark phases were mixed up. One solution is to reset the index after the warming phase but your suggestion is much better. I implemented. Thank you!
> 
> ```
> java -jar philadelphia-initiator.jar localhost 4000 50000 10000
>                                                    
> Warming up...
> Benchmarking...
> Results (n = 50000)
> 
>    50.00%:     361.22 µs
>    90.00%:  127139.84 µs
>    99.00%:  170917.89 µs
>    99.90%:  178388.99 µs
>    99.99%:  179568.64 µs
>   100.00%:  179961.86 µs
> ```

On what platform did you get these numbers? 😃 I get quite a bit more reasonable ones on my laptop:
```
philadelphia (bug-coordinated-omission) λ java -jar examples/initiator/philadelphia-initiator.jar localhost 4000 50000 10000
Warming up...
Benchmarking...
Results (n = 50000)

   50.00%:      19.18 µs
   90.00%:      37.60 µs
   99.00%:      53.57 µs
   99.90%:      80.19 µs
   99.99%:     222.46 µs
  100.00%:     318.72 µs
```

jvirtanen:
Thank you, @arturo-salinas! 🙌

Issue #95, opened 2019-01-25
Improve FIX message handling


jvirtanen:
Currently Philadelphia represents both incoming and outgoing FIX messages as instances of the [`FIXMessage`](https://github.com/paritytrading/philadelphia/blob/2.0.0/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java) class. A `FIXMessage` object is a sequential collection of tag-value pairs. Each tag is represented by an `int`, and each value is represented by a  [`FIXValue`](). `FIXValue` provides both read and write access to a value.

Internally, each `FIXValue` maintains an independent byte array as a backing store. On each read, the FIX value representation is parsed into an appropriate Java type, such as `int` or a  `double`. Similarly, on each write, a Java value is formatted into the corresponding FIX value representation. The size of the backing store is defined by [`FIXConfig#getFieldCapacity()`](https://github.com/paritytrading/philadelphia/blob/master/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConfig.java#L124).

When a message is parsed from a `ByteBuffer`, a `FIXMessage` instance essentially scatters the `ByteBuffer` contents into the `FIXValue` objects it holds. When a message is formatted into a `ByteBuffer`, the opposite takes place.

I suggest the following improvement to this architecture. A `FIXMessage` will maintain one contiguous byte array as a backing store. For outgoing messages, `FIXMessage` will provide methods that append a field of a specific type into the message. For example:
```java
/**
 * Add an integer field.
 *
 * @param tag the tag
 * @param value the value
 */
public void addInt(int tag, int value);
```
Appending a field will format the both the tag and the value directly into the backing store. This way, sending the message will only involve copying the backing store into `ByteBuffer`. (As a subsequent step, the backing store could be implemented directly as `ByteBuffer` to avoid the copy altogether.)

For incoming messages, `FIXMessage` will provide methods for accessing values of specific types. For example:
```java
/**
 * Get the integer value at the specified index.
 *
 * @param index the index
 * @return the integer value
 * @throws FIXValueFormatException if the value is not an integer
 */
public int intAt(int index);

/**
 * Get the integer value of the first instance of a field with the
 * specified tag.
 *
 * @param tag the tag
 * @return the integer value
 * @throws FIXValueFormatException if the value is not an integer
 */
public int intOf(int tag);
```
[...]

PR #107, opened 2019-06-19
Implement buffer direct allocation in FIXValue

arturo-salinas:
This pull request is meant for discussion and early feedback. 

I thought using a ByteBuffer (with direct allocation) as storage instead of an array could improve performance but my experiments show that the performance has actually decreased (see attached result files).
[output_test_original.txt](https://github.com/paritytrading/philadelphia/files/3308209/output_test_original.txt)
[output_test_bytebuffer.txt](https://github.com/paritytrading/philadelphia/files/3308210/output_test_bytebuffer.txt). 

However, it looks that the benchmark: asString has improved. 

In another experiment I noticed that direct allocation has actually decreased the performance further. 

Next steps could be to not use direct allocation and to do the ByteBuffer allocation in FIXMessage. But I will wait for feedback.

### DIFF
--- a/applications/client/src/main/java/com/paritytrading/philadelphia/client/Message.java
+++ b/applications/client/src/main/java/com/paritytrading/philadelphia/client/Message.java
@@ -90,7 +90,7 @@ class Message {
 
     void put(FIXMessage message) {
         for (Field field : fields)
-            message.addField(field.getTag()).setString(field.getValue());
+            message.addString(field.getTag(), field.getValue());
     }
 
     @Override
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
@@ -251,7 +251,7 @@ public class FIXConnection implements Closeable {
     public void prepare(FIXMessage message, char msgType) {
         message.reset();
 
-        message.addField(MsgType).setChar(msgType);
+        message.addChar(MsgType, msgType);
 
         prepare(message);
     }
@@ -266,16 +266,16 @@ public class FIXConnection implements Closeable {
     public void prepare(FIXMessage message, CharSequence msgType) {
         message.reset();
 
-        message.addField(MsgType).setString(msgType);
+        message.addString(MsgType, msgType);
 
         prepare(message);
     }
 
     private void prepare(FIXMessage message) {
-        message.addField(SenderCompID).setString(senderCompId);
-        message.addField(TargetCompID).setString(targetCompId);
-        message.addField(MsgSeqNum).setInt(txMsgSeqNum);
-        message.addField(SendingTime).setString(currentTimestamp);
+        message.addString(SenderCompID, senderCompId);
+        message.addString(TargetCompID, targetCompId);
+        message.addInt(MsgSeqNum, txMsgSeqNum);
+        message.addString(SendingTime, currentTimestamp);
     }
 
     /**
@@ -459,9 +459,9 @@ public class FIXConnection implements Closeable {
     public void sendReject(long refSeqNum, long sessionRejectReason, CharSequence text) throws IOException {
         prepare(txMessage, Reject);
 
-        txMessage.addField(RefSeqNum).setInt(refSeqNum);
-        txMessage.addField(SessionRejectReason).setInt(sessionRejectReason);
-        txMessage.addField(Text).setString(text);
+        txMessage.addInt(RefSeqNum, refSeqNum);
+        txMessage.addInt(SessionRejectReason, sessionRejectReason);
+        txMessage.addString(Text, text);
 
         send(txMessage);
     }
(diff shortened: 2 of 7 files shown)


arturo-salinas:
This pull request is meant for discussion and early feedback. 

I thought using a BufferBytes as storage instead of an array could improve performance but my experiments show that the performance has actually decreased (see attached result files).
[output_test_original.txt](https://github.com/paritytrading/philadelphia/files/3308209/output_test_original.txt)
[output_test_bytebuffer.txt](https://github.com/paritytrading/philadelphia/files/3308210/output_test_bytebuffer.txt). 

However, it looks that the benchmark: asString has improved.

jvirtanen:
Interesting experiments, @arturo-salinas! I will have a look. 🔍

jvirtanen:
I took a bit of time to think about this. 💡 

In the big picture, what we would like to optimize are the following paths:
- values are formatted into an outgoing message and the message is written to the network
- an incoming message is read from the network and values are parsed from it

Currently this is implemented as follows:
- when writing a value, e.g. with `FIXValue#setInt`, the FIX value representation is formatted into a byte array maintained by the `FIXValue` object
- when reading a value, e.g. with `FIXValue#asInt`, the FIX value representation is parsed from the byte array maintained by the `FIXValue` object
- when sending a message, the `FIXValue` objects maintained by the `FIXMessage` instance are sequentially written out into the (direct) transmission `ByteBuffer`
- when receiving a message, the `FIXValue` objects maintained by the `FIXMessage` instance are sequentially read in from the (direct) receive `ByteBuffer`

Now, having a byte array as a backing store for a `FIXValue` object is likely faster for formatting and parsing the FIX value representation. A non-direct `ByteBuffer` is essentially just a wrapper around a byte array, and a direct `ByteBuffer` is backed by memory outside of the Java heap. Typically bulk operations are faster with a direct `ByteBuffer` than a non-direct `ByteBuffer` whereas writing and reading single bytes tends to be slower. Finally, to actually write bytes to the network using NIO, they have to reside in a direct `ByteBuffer`. (If you don't put them into one by yourself, NIO will do this for you.)

(Note that many `FIXValue` methods are also currently under 35 bytes and inline well even on the C1 compiler. If the byte array is replaced by `ByteBuffer`, this is likely no longer the case.)

Also your performance test results seem to corroborate the above statements.

In the vein of what you mention in the pull request description, one very interesting thing to try out would be to use one contiguous byte array for all `FIXValue` objects within a `FIXMessage` instance. This byte array would be owned by the `FIXMessage` object, and each `FIXValue` instance would get a "slice" of it (an offset and a length).

What do you think, @arturo-salinas?

arturo-salinas:
I am continuing with the experiments. In the last commit I created a common ByteBuffer buffer which is allocated in FIXMessage. However, as the performance results show this is not a good idea. 

The problem with the parsing of incoming messages is that we are sequentially reading (getting) each byte of the direct ByteBuffer buffer where the message is placed. We cannot use buck operations because we do not know the length of each FIXValue (we do not know where the SOH mark is located). What I can try next is to use a buck copy from the direct ByteBuffer where the message is placed to a common ByteBuffer located in the heap which will be used for all the FIXValue objects. We can do that because we know the total length of the body. Then I think we can remove the array of int's which represents the tags and store the tag value inside FIXValue. At the moment of reading (getting) the FIX value, FIXValue will get a ByteBuffer slice of the common ByteBuffer which represents the FIX tag and FIX value. I hope this operation will be faster because it will be done just in the heap.

I also noticed that when the checksum is enabled; the sequential read of the direct ByteBuffer is done twice. What about to do the checksum at the same time when we are reading the FIX tag and FIX value, @jvirtanen? I guess the case when the checksum is not enabled will be a little bit slower because I would need to add a "if" each time a byte is read.

For the outgoing message I think it is durable to idea of the issue #95; we can replace the addField of FIXMessage for new methods such as addInt(tag, value), etc. Each of these methods will write to the common ByteBuffer (heap) directly and this will mean the "set" methods of FIXValue will be moved to FIXMessage. Then we can do a buck copy of the common ByteBuffer of FIXMessage to the direct ByteBuffer where the message is supposed to be sent. 

Any comments, @jvirtanen ?

arturo-salinas:
@jvirtanen I implemented a buffer for outgoing messages in FIXMessage. It looks that performance has been increased for FIXMessageBenchmark.formatAndPut.

arturo-salinas:
Here I attached the result of the performance tests:

[output_test_original.txt](https://github.com/paritytrading/philadelphia/files/3413513/output_test_original.txt)
[output_test_outgoing_new.txt](https://github.com/paritytrading/philadelphia/files/3413514/output_test_outgoing_new.txt)

ddimtirov:
Keep in mind that even a heap `ByteBuffer` has a lot of range checks which are not always optimized away (or take a long time and many optimization cycles to optimize). 

I'd recommend to keep a private reference to the HBB's backing array and use directly in performance sensitive spots such as parsing. Even better, introduce the notion of message pool - a class that allocates one large array of bytes and can create messages with backing byte buffers wrapping slices of it. This guarantees cache locality, and you can do fancy stuff to ensure alignment.

It is kind of run of the mill design, so you may want to look around to see if you can copy anything (i.e. out of the Aeron ecosystem)

jvirtanen:
Closing this for now.

PR #108, opened 2019-06-19
Fixvalue direct v2

arturo-salinas:
I noticed that FIXValue members: offset and length can be actually eliminated if ByteBuffer is used as underlying storage. However, my experiments show that it might not be a good idea. Or it can be there is some bugs in the implementation.
[output_no_offset_no_length.txt](https://github.com/paritytrading/philadelphia/files/3308232/output_no_offset_no_length.txt)


jvirtanen:
As discussed on #107, a non-direct `ByteBuffer` is essentially a wrapper on a byte array, and a direct `ByteBuffer` is backed by memory outside the Java heap. Due to the added overhead, operating on individual bytes in either data structure is slower than operating on a byte array directly.

In the long term, an approach closer to #95 and #107 is likely going to be the future direction for this library.

PR #109, opened 2019-06-22
philadelphia-core: Improve MsgType(35) handling

jvirtanen:
When checking for SequenceReset(4), check the first byte of the MsgType(35) value before checking its length. This avoids one conditional check for most messages.

Note that the length of the MsgType(35) value could in principle be zero. In that case, "the first byte of the value" is actually the SOH character.

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
@@ -532,7 +532,7 @@ public class FIXConnection implements Closeable {
                 return;
             }
 
-            if (msgType.length() == 1 && msgType.byteAt(0) == SequenceReset) {
+            if (msgType.byteAt(0) == SequenceReset && msgType.length() == 1) {
                 if (handleSequenceReset(message))
                     return;
             }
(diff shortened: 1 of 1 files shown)


jheusser:
Good change! 

Two comments/questions on this method:
- `msgType.length() == 1`: this comparison has to be done at least [once](https://github.com/paritytrading/philadelphia/blob/d261d0e2be438ecb4cd15d9a6d266695f0c98551/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java#L547), maybe twice. Would it make sense to factor it out as boolean?
-  The method `handleSequenceReset` is also invoked as a case in the switch statement, would this ever actually happen (especially as `rxMsgSeqNum` is increased by then) or are all sequence resets caught in the first if statement which this PR touches? If so, it could be removed from the switch statement.

jvirtanen:
> * `msgType.length() == 1`: this comparison has to be done at least [once](https://github.com/paritytrading/philadelphia/blob/d261d0e2be438ecb4cd15d9a6d266695f0c98551/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java#L547), maybe twice. Would it make sense to factor it out as boolean?

Possibly! I think this is something that could be looked into as a further improvement. Perhaps also e.g. extracting the first byte of the MsgType(35) value could be useful, it's also accessed a few times.

>  * The method `handleSequenceReset` is also invoked as a case in the switch statement, would this ever actually happen (especially as `rxMsgSeqNum` is increased by then) or are all sequence resets caught in the first if statement which this PR touches? If so, it could be removed from the switch statement.

No, it wouldn't ever happen. Good catch, I removed the superfluous case from the switch statement. 👍

PR #115, opened 2019-07-08
philadelphia-core: Improve 'FIXValue#asFloat' and 'FIXValue#setFloat'

jvirtanen:


libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
```diff
@@ -211,48 +232,87 @@ public void setInt(long x) {
...
 
-            x = 10 * x + b - '0';
+            significand = 10 * significand + b - '0';
 
-            factor *= 10;
+            exponent++;
         }
 
-        return sign * (factor > 0.0 ? x / factor : x);
+        double x = exponent != 0 ? significand / POWERS_OF_TEN[exponent] : significand;
```
jheusser: Could you drop this conditional and just write below as `POWERS_OF_TEN[0] == 1`:
```
double x = significand / POWERS_OF_TEN[exponent];
```

Maybe something for the benchmark to decide.

jvirtanen: Good catch! 🔍  We certainly don't want to optimize for `exponent == 0` at the expense of `exponent != 0`.

jvirtanen: This is now fixed. 👍

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
```diff
@@ -211,48 +232,87 @@ public void setInt(long x) {
...
+            significand = 10 * significand + b - '0';
 
-            factor *= 10;
+            exponent++;
         }
 
-        return sign * (factor > 0.0 ? x / factor : x);
+        double x = significand / POWERS_OF_TEN[exponent];
+
+        return negative ? -x : +x;
```
jheusser: What about keeping this branchless?

jvirtanen: I don't see a perceptible difference between this and a branchless version, but then again the performance test only tests positive numbers. 🙂 On the other hand, I can imagine many applications will generally also predominantly if not exclusively work with positive numbers.

In the future, I will look into some related enhancements in the performance tests: positive and negative numbers, numbers with different number of digits and so on.

PR #116, opened 2019-07-08
Polish -- close local connection

JanStureNielsen:


JanStureNielsen:
The CI failure appears to be an environment configuration issue, unrelated to the proposed change.

jvirtanen:
Hi @JanStureNielsen! The CI failure _is_ actually caused by the proposed change. 🙂  `FIXConnection` is `Closeable`, but the invocation of `FIXConnection#close()` fails in this test case with a `NullPointerException` because the `channel` argument passed to the constructor is `null`.

I think the current behavior is okay, because the purpose of the test case is just to confirm that a `FIXConnection` instance can be created with `FIXVersion.FIXT_1_1`. Does your IDE or some other tool warn about a potential resource leak in this test case?

JanStureNielsen:
Makes sense, @jvirtanen. STS displays a resource leak warning; other tools will likely do the same. The warning could be suppressed which might help communicate the intent a bit more as well. Happy to close the PR or modify accordingly.

jvirtanen:
> Makes sense, @jvirtanen. STS displays a resource leak warning; other tools will likely do the same. The warning could be suppressed which might help communicate the intent a bit more as well. Happy to close the PR or modify accordingly.

Yes, I think suppressing the warning is a good idea. 👍  I presume `@SuppressWarnings("resource")` should do it?

jvirtanen:
Thank you, @JanStureNielsen! 🙌

PR #126, opened 2020-02-17

jvirtanen:
Inlining parsing of BeginString(8), BodyLength(9), and CheckSum(10) tags makes message parsing 5.9% faster in the performance test when the incoming CheckSum(10) check is disabled and 6.1% faster when it is enabled.


jheusser:
So it's faster _with_ running the checksum than without?

jvirtanen:
> So it's faster _with_ running the checksum than without?

No, the impact is just bigger with the incoming CheckSum(10) check. This makes sense, as one of the three tags, CheckSum(10), is only checked with the incoming CheckSum(10) check enabled. (So for that case, in total three `FIXTags.get()` call sites are eliminated whereas with the check enabled only two.) 🙂

PR #128, opened 2020-02-24
philadelphia-client: Fix 'wait' command

### LINKED ISSUE #117: philadelphia-client: Fix 'wait' command
The `wait` command is broken: it does not look at all received messages since being issued but just the last received message after each wait interval. This means that if `wait A` has been invoked and two messages with MsgType(35) values A and B, respectively, are received within one wait interval, the command only sees the MsgType(35) value B of the second message and continues waiting.

Fix the `wait` command so that it looks at all received messages since being issued.

jvirtanen:
The `wait` command looks at received messages at an interval, stopping if the last received message has a matching MsgType(35) value. If two or more messages are received within one interval, the `wait` command may miss a matching MsgType(35) value and erroneously continue waiting.

Fix the `wait` command so that it looks at all received messages since being issued.

Closes #117.

### DIFF
--- a/applications/client/src/main/java/com/paritytrading/philadelphia/client/WaitCommand.java
+++ b/applications/client/src/main/java/com/paritytrading/philadelphia/client/WaitCommand.java
@@ -15,6 +15,8 @@
  */
 package com.paritytrading.philadelphia.client;
 
+import static java.util.stream.Collectors.*;
+
 import java.util.List;
 import java.util.Scanner;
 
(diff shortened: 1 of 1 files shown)


jvirtanen:
I have tested this locally by updating the [Parity FIX Gateway test suite](https://github.com/paritytrading/parity/blob/master/applications/fix/test/test.txt) to use `wait` instead of `sleep`.

PR #132, opened 2020-02-25
Configure GitHub Actions

jvirtanen:


jvirtanen:
This pull request replaces Travis with GitHub Actions. You can see the [latest results of that in my fork](https://github.com/jvirtanen/philadelphia/actions/runs/45377403). The initial motivation is to reduce moving parts; perhaps later GitHub Actions could be used for automating the release process as well.

PR #139, opened 2020-02-27
philadelphia-core: Remove transitive Joda-Time dependency

jvirtanen:
Currently, Philadelphia Core depends on Joda-Time 2.10.5. This means that any application that has a dependency on Philadelphia Core also has a transitive dependency on Joda-Time 2.10.5.

Replace the transitive Joda-Time 2.10.5 dependency with a Joda-Time 2.x dependency. This makes it possible for applications to manage their Joda-Time 2.x and Philadelphia Core dependencies independently.

This also means that applications that use Philadelphia Core but do not yet use Joda-Time 2.x need to add a dependency on Joda-Time 2.x as well.

### DIFF
--- a/applications/client/pom.xml
+++ b/applications/client/pom.xml
@@ -39,6 +39,10 @@
       <artifactId>config</artifactId>
     </dependency>
     <dependency>
+      <groupId>joda-time</groupId>
+      <artifactId>joda-time</artifactId>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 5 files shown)


jvirtanen:
I have tested this by compiling Philadelphia Core against Joda-Time 2.10.5 and running Philadelphia Terminal Client with Joda-Time 2.9.9.

PR #143, opened 2020-02-29
philadelphia-core: Improve 'FIXValue#asTimeOnly'

jvirtanen:
Set the fields all at once rather than one by one. This makes the method 33% faster.


jvirtanen:
I noticed this while working on #141 and #142.

PR #151, opened 2020-03-23
philadelphia-core: Improve backing store management

jvirtanen:
Start using shared backing arrays for value containers.

Allocate one or more backing arrays, called _chunks_, per message container. Introduce a new configuration option, `chunkSize`, to specify the size of one such backing array.

In order to enable this, first make it possible to specify a backing array for a value container.

Add two new internal attributes: `start` and `end`. These are the start index (inclusive) and the end index (exclusive) for the value container's slot in the backing array.

The existing internal attributes `offset` and `length` retain the same semantics as before. In particular, `offset` still refers to the position of the contained value within the entire backing array  and is not relative to `start`.

Because `offset` and `length` work the same way as before, the value accessors, such as `FIXValue#asBoolean`, do not change. However, the value mutators, such as `FIXValue#setBoolean` do change: the value container's slot in the backing array now starts at `start` instead of 0 and ends at `end` instead of `bytes.length`.

The bytecode changes are as follows:

Method | Before | After | Change
-|-|-|-
`set` | 41 | 41 | 0
`setBoolean` | 34 | 45 | +11
`setChar` | 25 | 36 | +11
`setInt` | 102 | 99 | -3
`setFloat` | 177 | 174 | -3
`setString` | 55 | 68 | +13
`setDate` | 59 | 80 | +21
`setTimeOnly` | 119 | 165 | +46
`setTimestamp` | 172 | 238 | +66
`setCheckSum` | 28 | 39 | +11
`get` | 65 | 67 | +2

The hope is that improved data locality will produce performance gains despite the overhead of increased code complexity. However, detailed performance testing is still to be done.

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConfig.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConfig.java
@@ -30,7 +30,7 @@ public class FIXConfig {
     private final long       outgoingMsgSeqNum;
     private final int        maxFieldCount;
     private final int        fieldCapacity;
-    private final int        chunkSize;
+    private final int        blockSize;
     private final int        rxBufferCapacity;
     private final int        txBufferCapacity;
     private final boolean    checkSumEnabled;
@@ -46,7 +46,7 @@ public class FIXConfig {
      * @param outgoingMsgSeqNum the outgoing MsgSeqNum(34)
      * @param maxFieldCount the maximum number of fields in a message
      * @param fieldCapacity the field capacity
-     * @param chunkSize the backing store chunk size
+     * @param blockSize the backing store block size
      * @param rxBufferCapacity the receive buffer capacity
      * @param txBufferCapacity the transmit buffer capacity
      * @param checkSumEnabled the incoming CheckSum(10) check status
@@ -55,7 +55,7 @@ public class FIXConfig {
     public FIXConfig(FIXVersion version, String senderCompId,
             String targetCompId, int heartBtInt,
             long incomingMsgSeqNum, long outgoingMsgSeqNum,
-            int maxFieldCount, int fieldCapacity, int chunkSize,
+            int maxFieldCount, int fieldCapacity, int blockSize,
             int rxBufferCapacity, int txBufferCapacity,
             boolean checkSumEnabled) {
         this.version           = version;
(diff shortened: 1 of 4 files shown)


jheusser:
Looks good and nice that it did not change the API much. 

What's the purpose of the `chunkSize`?  Just asking this because the `FIXConfig` is already quite complex, maybe the total size could be deduced from the other fields such as `maxFieldCount` etc.

jvirtanen:
> What's the purpose of the `chunkSize`? Just asking this because the `FIXConfig` is already quite complex, maybe the total size could be deduced from the other fields such as `maxFieldCount` etc.

It is the size of a single backing array for a number of value containers. It is by default set to 4096 bytes (the page size). Depending on the maximum number of fields as well as the field capacity, one message container, upon construction, allocates one or more chunks for its value containers. You are right in that it would be nice to avoid adding yet another configuration parameter. 🙂

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConfig.java
```diff
@@ -234,6 +249,7 @@ public Builder() {
             outgoingMsgSeqNum = 1;
             maxFieldCount     = 64;
             fieldCapacity     = 64;
+            blockSize         = 4096;
```
penberg, line 252: The memory layout of Java arrays in memory is as follows: (1) object header (16 bytes) and (2) the array elements.

IOW, if you want to be fancy and micro-optimize, make the `blockSize` be 4080 by default, and your array object will fit in one 4 KiB hardware page:

```java
package parity;

import org.openjdk.jol.info.ClassLayout;

public class MemoryLayout {
  public static void main(String[] args) {
    System.out.println(ClassLayout.parseInstance(new byte[4096]).toPrintable());
    System.out.println(ClassLayout.parseInstance(new byte[4080]).toPrintable());
  }
}
```

```console
# WARNING: Unable to get Instrumentation. Dynamic Attach failed. You may add this JAR as -javaagent manually, or supply -Djdk.attach.allowAttachSelf
[B object internals:
OFF  SZ   TYPE DESCRIPTION               VALUE
  0   8        (object header: mark)     0x0000000000000001 (non-biasable; age: 0)
  8   4        (object header: class)    0x00076862
 12   4        (array length)            4096
 12   4        (alignment/padding gap)
 16 4096   byte [B.<elements>             N/A
Instance size: 4112 bytes
Space losses: 4 bytes internal + 0 bytes external = 4 bytes total

[B object internals:
OFF  SZ   TYPE DESCRIPTION               VALUE
  0   8        (object header: mark)     0x0000000000000001 (non-biasable; age: 0)
  8   4        (object header: class)    0x00076862
 12   4        (array length)            4080
 12   4        (alignment/padding gap)
 16 4080   byte [B.<elements>             N/A
Instance size: 4096 bytes
Space losses: 4 bytes internal + 0 bytes external = 4 bytes total
```

jvirtanen: This is a good point, @penberg! 👍🏼  I'm actually in the process of rebasing this pull request on `master` again; some of the `FIXValue` internals have got changed since the spring of 2020.

jvirtanen:
Closing this for now.

PR #153, opened 2020-04-30
philadelphia-core: Get first tag or value starting from given index

ksmr:
Extend the existing functionality of `FIXMessage#valueOf` and `FIXMessage#indexOf` with a second argument specifying the index to start looking from.

This is useful when a message has the same tag multiple times with different values and one wishes to cycle through them.

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java
@@ -139,6 +139,24 @@ public class FIXMessage {
     }
 
     /**
+     * Get the value container of the first instance of a field with the
+     * specified tag, starting at the specified index.
+     *
+     * @param tag the tag
+     * @param fromIndex the index to start the search from
+     * @return the value container or {@code null} if there are no instances
+     *   of a field with the specified tag
+     */
+    public FIXValue valueOf(int tag, int fromIndex) {
+        for (; fromIndex < count; fromIndex++) {
+            if (tags[fromIndex] == tag)
+                return values[fromIndex];
+        }
+
+        return null;
+    }
+
+    /**
      * Get the index of the first instance of a field with the specified tag.
      *
      * @param tag the tag
@@ -149,6 +167,24 @@ public class FIXMessage {
         for (int i = 0; i < count; i++) {
             if (tags[i] == tag)
                 return i;
+        }
+
+        return -1;
+    }
+
+    /**
+     * Get the index of the first instance of a field with the specified tag,
+     * starting at the specified index.
+     *
+     * @param tag the tag
+     * @param fromIndex the index to start the search from
+     * @return the index or -1 if there are no instances of a field with the
+     *   specified tag
+     */
+    public int indexOf(int tag, int fromIndex) {
+        for (; fromIndex < count; fromIndex++) {
+            if (tags[fromIndex] == tag)
+                return fromIndex;
         }
 
         return -1;


libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java
```diff
@@ -130,7 +130,20 @@ public void setFieldCount(int count) {
      *   of a field with the specified tag
      */
     public FIXValue valueOf(int tag) {
-        for (int i = 0; i < count; i++) {
+        return valueOf(tag, 0);
+    }
```
jvirtanen, line 134: I would retain the existing `valueOf` body to avoid the indirection in the most common case.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java
```diff
@@ -146,7 +159,19 @@ public FIXValue valueOf(int tag) {
      *   specified tag
      */
     public int indexOf(int tag) {
-        for (int i = 0; i < count; i++) {
+        return indexOf(tag, 0);
+    }
```
jvirtanen, line 163: I would retain the existing `indexOf` body to to avoid indirection in the most common case.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java
```diff
@@ -130,7 +130,20 @@ public void setFieldCount(int count) {
...
-        for (int i = 0; i < count; i++) {
+        return valueOf(tag, 0);
+    }
+
+    /**
+     * Get the value container of the first instance of a field with the
+     * specified tag, starting at the specified index.
+     *
+     * @param tag the tag
+     * @param fromIndex the index to start the search from.
```
jvirtanen, line 141: Let's remove the full stop from the end of the `@param` tag.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java
```diff
@@ -146,7 +159,19 @@ public FIXValue valueOf(int tag) {
...
     public int indexOf(int tag) {
-        for (int i = 0; i < count; i++) {
+        return indexOf(tag, 0);
+    }
+
+    /**
+     * Get the index of the first instance of a field with the specified tag, starting at the specified index.
+     *
+     * @param tag the tag
+     * @param fromIndex the index to start the search from.
```
jvirtanen, line 169: Let's remove the full stop from the end of the `@param` tag.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java
```diff
@@ -130,7 +130,20 @@ public void setFieldCount(int count) {
...
+     * Get the value container of the first instance of a field with the
+     * specified tag, starting at the specified index.
+     *
+     * @param tag the tag
+     * @param fromIndex the index to start the search from.
+     * @return the value container or {@code null} if there are no instances
+     *   of a field with the specified tag
+     */
+    public FIXValue valueOf(int tag, int fromIndex) {
+        for (int i = fromIndex; i < count; i++) {
```
jvirtanen, line 146: In the `#valueOf(int,int)` method, we could simplify the resulting bytecode slightly by eliminating `i` and using `fromIndex` as the index variable within the loop. As in:
```java
for (; fromIndex < count; fromIndex++) {
    ...
```

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java
```diff
@@ -146,7 +159,19 @@ public FIXValue valueOf(int tag) {
...
+    /**
+     * Get the index of the first instance of a field with the specified tag, starting at the specified index.
+     *
+     * @param tag the tag
+     * @param fromIndex the index to start the search from.
+     * @return the index or -1 if there are no instances of a field with the
+     *   specified tag
+     */
+    public int indexOf(int tag, int fromIndex) {
+        for (int i = fromIndex; i < count; i++) {
```
jvirtanen, line 174: In the `#indexOf(int,int)` method, we could simplify the resulting bytecode slightly by eliminating `i` and using `fromIndex` as the index variable within the loop. As in:
```java
for (; fromIndex < count; fromIndex++) {
    ...
```

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessage.java
```diff
@@ -146,7 +159,19 @@ public FIXValue valueOf(int tag) {
      *   specified tag
      */
     public int indexOf(int tag) {
-        for (int i = 0; i < count; i++) {
+        return indexOf(tag, 0);
+    }
+
+    /**
+     * Get the index of the first instance of a field with the specified tag, starting at the specified index.
```
jvirtanen, line 166: This could also use word wrap at around 80 characters. 🤓

jvirtanen:
Thank you, @ksmr, great work! 🙌

PR #159, opened 2020-08-23
philadelphia-core: Add 'FIXValue#asString(Appendable)'

### LINKED ISSUE #158: Add 'FIXValue#asString(Appendable)'
Check from Java Language Specification and Java Virtual Machine Specification for Java 8 if it is possibly to widen the parameter type of [`FIXValue#asString(StringBuilder)`](https://github.com/paritytrading/philadelphia/blob/master/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java#L355) from [`StringBuilder`](https://docs.oracle.com/javase/8/docs/api/java/lang/StringBuilder.html) to [`Appendable`](https://docs.oracle.com/javase/8/docs/api/java/lang/Appendable.html) without breaking binary compatibility. If it is, do so. Otherwise, add `FIXValue#asString(Appendable)` beside the existing method.

jvirtanen:
Closes #158.

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
@@ -18,6 +18,7 @@ package com.paritytrading.philadelphia;
 import static com.paritytrading.philadelphia.FIX.*;
 import static java.nio.charset.StandardCharsets.*;
 
+import java.io.IOException;
 import java.nio.BufferOverflowException;
 import java.nio.ByteBuffer;
 import java.nio.ReadOnlyBufferException;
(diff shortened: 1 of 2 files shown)


libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
```diff
@@ -346,6 +347,18 @@ public String asString() {
...
     }
 
+    /**
+     * Get the value as a string. The value is appended to the provided
+     * appendable.
+     *
+     * @param x an appendable
+     * @throws IOException if an I/O error occurs
+     */
+    public void asString(Appendable x) throws IOException {
```
jheusser, line 357: Would it make sense to remove the method `asString(StringBuilder x)` now, or not due to backwards compatibility?

jheusser (approved):
Nice! 💯

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
jvirtanen: Removing that method would introduce a binary incompatible change. Also, it's worth noting that `Appendable#append` may throw an `IOException` whereas `StringBuilder#append` may not.

PR #160, opened 2020-08-23
philadelphia-core: Add 'FIXValue#contentEquals'

### LINKED ISSUE #157: Add 'FIXValue#contentEquals' methods
Add the following methods to the `FIXValue` class:

```java
boolean contentEquals(byte x);

boolean contentEquals(byte[] x);

boolean contentEquals(char x);

boolean contentEquals(CharSequence x);
```

These methods are inspired by [the method by the same name](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#contentEquals-java.lang.CharSequence-) in `String` and can be helpful especially when working with enumerations that mix `char` and `String` members, such as [MsgType(35)](https://github.com/paritytrading/philadelphia/blob/master/libraries/fix50sp2/src/main/java/com/paritytrading/philadelphia/fix50sp2/FIX50SP2MsgTypes.java).

jvirtanen:
Closes #157.

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
@@ -92,6 +92,74 @@ public class FIXValue {
     }
 
     /**
+     * Compare the string representation of this value to the specified
+     * character.
+     *
+     * @param x the character to compare the string representation of this
+     *   value against
+     * @return true if this value represents the same string as the specified
+     *   character, otherwise false
+     */
+    public boolean contentEquals(byte x) {
+        return length == 1 && bytes[offset] == x;
+    }
+
+    /**
+     * Compare the string representation of this value to the specified
+     * character sequence.
+     *
+     * @param x the character sequence to compare the string representation of
+     *   this value against
+     * @return true if this value represents the same string as the specified
+     *   character sequence, otherwise false
+     */
+    public boolean contentEquals(byte[] x) {
+        if (length != x.length)
+            return false;
+
+        for (int i = 0; i < length; i++) {
+            if (bytes[offset + i] != x[i])
+                return false;
+        }
+
+        return true;
+    }
+
+    /**
+     * Compare the string representation of this value to the specified
+     * character.
+     *
+     * @param x the character to compare the string representation of this
+     *   value against
+     * @return true if this value represents the same string as the specified
+     *   character, otherwise false
+     */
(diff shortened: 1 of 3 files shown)


libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
```diff
@@ -91,6 +91,74 @@ public int length() {
...
+     * @param x the character sequence to compare the string representation of
+     *   this value against
+     * @return true if this value represents the same string as the specified
+     *   character sequence, otherwise false
+     */
+    public boolean contentEquals(byte[] x) {
+        if (length != x.length)
+            return false;
+
+        for (int i = 0; i < length; i++) {
```
jheusser, line 120: What about `Arrays.equals`? Also possibly an intrinsic in JDK9.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
```diff
@@ -91,6 +91,74 @@ public int length() {
...
+     * Compare the string representation of this value to the specified
+     * character.
+     *
+     * @param x the character to compare the string representation of this
+     *   value against
+     * @return true if this value represents the same string as the specified
+     *   character, otherwise false
+     */
+    public boolean contentEquals(byte x) {
+        return length == 1 && bytes[offset] == x;
```
jheusser, line 104: Why does this use `offset` while others don't?

jheusser (approved):
Nice! 💯

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
jvirtanen: Good catch! 🔍 They should all use `offset`.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
jvirtanen: Unfortunately that cannot be used, because `bytes` is always going to be longer by at least one element (the SOH). Therefore the actual arrays are not equal.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
jvirtanen: I fixed these, and I also improved the test cases so that they exercise situations in which `offset != 0`.

PR #162, opened 2020-11-06
Add FIX Latest support

jvirtanen:
[FIX Latest](https://www.fixtrading.org/online-specification/) is the latest FIX protocol version, currently based on FIX 5.0 SP2 EP258. It is specified in a new format called [FIX Orchestra](https://www.fixtrading.org/standards/fix-orchestra/). This pull request adds FIX Orchestra as an input source for Philadelphia Code Generator, the infrastructure to generate FIX Latest, as well as current FIX Latest itself.


applications/generate/philadelphia/orchestra.py
```diff
@@ -0,0 +1,129 @@
...
+
+def _has_values(code_set):
+    return not code_set.id_ in _NO_VALUES and not code_set.type_ == 'Boolean'
+
+
+_NAME_REPLACEMENTS = {
+    ('1779', '1'): 'Int',
+    ('1779', '6'): 'Float',
+    ('1779', '12'): 'Char',
+}
```
jvirtanen, line 76: I have opened [an issue on the FIX Trading Community repository](https://github.com/FIXTradingCommunity/orchestrations/issues/15) regarding the handling of these values.

jvirtanen:
I opened [a second issue on the FIX Trading Community repository](https://github.com/FIXTradingCommunity/orchestrations/issues/16) regarding the QuoteCondition(276) value "f " (FastMarketETH). The input source for FIX Repository, used for FIX protocol versions prior to FIX Latest, strips the trailing space character from that value. However, this is a good opportunity to check with FIX Trading Community whether that makes sense. (Some other FIX protocol implementations also strip it, some others do not.)

jvirtanen:
> I opened [a second issue on the FIX Trading Community repository](https://github.com/FIXTradingCommunity/orchestrations/issues/16) regarding the QuoteCondition(276) value "f " (FastMarketETH). The input source for FIX Repository, used for FIX protocol versions prior to FIX Latest, strips the trailing space character from that value. However, this is a good opportunity to check with FIX Trading Community whether that makes sense. (Some other FIX protocol implementations also strip it, some others do not.)

This [is an error](https://github.com/FIXTradingCommunity/orchestrations/issues/16#issuecomment-724576477), so also the input source for FIX Orchestra will strip the trailing space character from that value.

applications/generate/philadelphia/orchestra.py
jvirtanen: This feels like the best way to handle these for now.

PR #163, opened 2020-11-11
philadelphia-core: Improve buffer capacity management

jvirtanen:
Add `FIXConfig#getBufferCapacity()` and deprecate the following methods:

  - `FIXConfig#getRxBufferCapacity()`
  - `FIXConfig#getTxBufferCapacity()`

The purpose of this change is to simplify configuration. As the maximum number of fields and the field capacity already apply to both incoming and outgoing messages, it makes sense to have one buffer capacity that applies to both as well.

The deprecated methods will continue working as before until they get removed in Philadelphia 2.0.0.

### DIFF
--- a/applications/client/src/main/java/com/paritytrading/philadelphia/client/TerminalClient.java
+++ b/applications/client/src/main/java/com/paritytrading/philadelphia/client/TerminalClient.java
@@ -204,8 +204,7 @@ class TerminalClient implements Closeable {
             .setHeartBtInt(heartBtInt)
             .setMaxFieldCount(1024)
             .setFieldCapacity(1024)
-            .setRxBufferCapacity(1024 * 1024)
-            .setTxBufferCapacity(1024 * 1024);
+            .setBufferCapacity(1024 * 1024);
 
         TerminalClient.open(new InetSocketAddress(address, port), builder.build()).run(lines);
     }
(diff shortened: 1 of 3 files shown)


jvirtanen:
This is the first of my suggested configuration improvements. Both concern buffer capacity management; the second will introduce logic to automatically calculate the buffer capacity (unless explicitly set) based on the maximum number of fields and the field capacity, further simplifying configuration.

jheusser (approved):
Nice! 💯

Makes sense to simplify the configuration. Could the buffer capacity also just be computed since misconfiguring this can probably lead to weird results?

Also just one comment is that TX and RX messages can be quite different in size.

jvirtanen:
> Makes sense to simplify the configuration. Could the buffer capacity also just be computed since misconfiguring this can probably lead to weird results?

I agree, we should be able to do better here. What do you think about having something analogous to e.g. this more forgiving and dynamic approach in [another library](https://github.com/fasterjsonorg/jargon) I've been working on?
```java
/**
 * Get the minimum capacity for JSON nodes. The JSON tree preallocates
 * storage for handling a number of JSON nodes less than or equal to
 * this value.
 *
 * @return the minimum capacity for JSON nodes
 */
public int getMinNodeCapacity() {
    return minNodeCapacity;
}

/**
 * Get the maximum capacity for JSON nodes.
 * {@link JsonTree#reset(JsonParser)} throws a {@link JsonMappingException}
 * if the number of JSON nodes exceeds this value.
 *
 * @return the maximum capacity for JSON nodes
 */
public int getMaxNodeCapacity() {
    return maxNodeCapacity;
}
```

PR #164, opened 2020-11-24
philadelphia-core: Optimize 'FIXValue#setInt' and 'FIXValue#setFloat'

jvirtanen:
Optimize handling of positive numbers at the expense of negative numbers. This decreases the bytecode sizes of the methods as follows:

Method | Before | After
-|-|-
`setInt` | 102 | 79
`setFloat` | 177 | 161

The impact on the microbenchmarks is negligible.

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXValue.java
@@ -399,32 +399,32 @@ public class FIXValue {
      * @see #asFloat
      */
     public void setFloat(double x, int decimals) {
-        bytes[end - 1] = SOH;
+        int i = end;
+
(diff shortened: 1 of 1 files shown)


jheusser (approved):
Nice! 💯

Is there enough test coverage on negative and maybe 0 prices?

jvirtanen:
> Is there enough test coverage on negative and maybe 0 prices?

Good point. There are _some_, but I will improve the test coverage. For example, it looks like the current rounding half away from zero is actually not tested at all on the negative side.

jvirtanen:
This actually fixes `FIXValue#setInt()` for the value `Long.MIN_VALUE`, which has [special handling for `Math.abs()`](https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#abs-long-).

PR #167, opened 2020-12-08

jvirtanen:
Inline parsing of the BeginString(8) value. This makes message parsing 5.3% faster in the performance test when the incoming CheckSum(10) check is disabled and 4.1% faster when it is enabled.


jheusser (approved):
That is nice! 💯

PR #174, opened 2020-12-10

jvirtanen:
Optimize parsing of the BodyLength(9) and CheckSum(10) values. This makes message parsing 1.7% faster in the performance test when the incoming CheckSum(10) check is disabled and 7.6% faster when it is enabled.


jheusser (approved):
Very nice! 💯 

How much smaller can the parser get? 😄

jvirtanen:
> How much smaller can the parser get? 😄

I think we can definitely still do better than this. 🙂 Some of the logic for skipping garbled messages in the parser originates from the FIX protocol specification. It would make sense to either have a fast path to bypass that logic whenever possible or to make it completely optional.

PR #176, opened 2020-12-18
philadelphia-core: Improve 'FIXValue#toString'

jvirtanen:
Remove the pipe character (`|`) from the `FIXValue#toString()` return value. As this means that the return value of this method will be equal to the return value of the `FIXValue#asString()` method, remove the `FIXValue#toString(StringBuilder)` method as it becomes effectively a duplicate of the `FIXValue#asString(StringBuilder)` method.

Philadelphia uses the pipe character as a human-readable stand-in for the non-printable SOH character, which terminates a value in the FIX protocol. Although it is a part of the internal data representation in the `FIXValue` class, I feel that aligning the `toString` method to the `asString` method makes much more sense for most use cases of this method.

As this method was added after the most recent release, 1.2.0, this change will have no impact on existing Philadelphia applications.


jheusser (approved):
Looks good to me! 👍🏽

PR #179, opened 2021-03-25
philadelphia-core: Remove 'FIXTimestamps.setDigits' optimization

jvirtanen:
Although the original change decreases the bytecode size of the method from 36 to 28 bytes, which is within HotSpot's default maximum inline size of 35 bytes, this appears to cause a 1-2% performance regression when HotSpot has fully optimized this code. Fix this by removing the bytecode optimization.

This reverts commit 1571bf60bbba825fd216794a3d09bb0f41092a29.


jheusser (approved):
Nice! 💯

PR #181, opened 2021-03-29
philadelphia-core: Remove 'FIXValue#setDigits' optimization

jvirtanen:
Although the original change decreases the bytecode size of the method from 39 to 31 bytes, which is within HotSpot's default maximum inline size of 35 bytes, this appears to cause a 1-2% performance regression when HotSpot has fully optimized this code. Fix this by removing the bytecode optimization.

This reverts commit 64c68f47d57e833613a531cc7429a6d59ec23f3a.


jheusser (approved):
Nice! 💯

PR #186, opened 2021-05-17
scripts: Fix 'format-perf-test' script

jvirtanen:
Embarrassingly, the baseline and candidate results are mixed up. Fix this.


jheusser (approved):
Good catch! 💥

PR #193, opened 2021-05-26
philadelphia-core: Fix 'FIXMessageOverflowException' documentation

jvirtanen:
This exception is thrown both when the message length exceeds the capacity of the receive buffer as well as when the number of fields exceeds the capacity of a message container.


jheusser (approved):
Great! 👍

PR #196, opened 2021-08-16
philadelphia-core: Use 'FIXValue' for current timestamp

jvirtanen:
As `FIXValue#setTimestamp()` is faster than `FIXTimestamps#append()` and `FIXValue` now implements the `CharSequence` interface, prefer `FIXValue` to `StringBuilder` for the current timestamp.


jheusser (approved):
Looks good to me! 👍

PR #197, opened 2021-08-16
philadelphia-fixlatest: Update to EP269

jvirtanen:


libraries/fixlatest/src/main/java/com/paritytrading/philadelphia/fixlatest/FIXLatestEnumerations.java
```diff
@@ -7779,21 +7741,6 @@ private RiskLimitRequestResultValues() {
...
-
-        public static final int Accepted            = 0;
-        public static final int AcceptedWithChanges = 1;
-        public static final int Rejected            = 2;
-        public static final int AcceptancePending   = 3;
-
-        private RiskLimitRequestStatusValues() {
-        }
-
-    }
```
jvirtanen, line 7795: RiskLimitRequestStatus(1762), among others, now shares its value range with PartyDetailRequestStatus(1878). It would be good to document these kind of dependencies for all dialects based on FIX Orchestra.

jheusser (approved):
Nice! 💯

jvirtanen:
I will implement an improvement into the code generation before merging this. This impacts primarily or possibly only the FIX Latest dialect, which will gain better Javadoc documentation for many enumerations. (At least in FIX Latest, many enumerations are used not only as the value range for a single field but multiple related fields.)

jvirtanen:
Rebased on `master`.

PR #207, opened 2021-09-18
philadelphia-core: Remove 'Clock'

jvirtanen:
Instead of passing `FIXConnection` a time source and asking it to update the current time, pass it the current time itself. Rename the `updateCurrentTimestamp` method to `setCurrentTimeMillis`, and add a `getCurrentTimeMillis` method to complement it.


jheusser (approved):
This makes sense to me 👍🏽

PR #208, opened 2021-09-21
philadelphia-core: Add 'FIXConnection#reprepare()'

jvirtanen:
Replace the existing `updateCompID()` method with one that feels more at home with the existing `prepare()` and `update()` methods.

### DIFF
--- a/examples/acceptor/src/main/java/com/paritytrading/philadelphia/acceptor/Session.java
+++ b/examples/acceptor/src/main/java/com/paritytrading/philadelphia/acceptor/Session.java
@@ -77,7 +77,7 @@ class Session implements FIXMessageListener {
             public void logon(FIXConnection connection, FIXMessage message) throws IOException {
                 connection.sendLogon(true);
 
-                connection.updateCompID(report);
+                connection.reprepare(report);
             }
 
             @Override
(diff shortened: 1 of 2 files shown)


examples/acceptor/src/main/java/com/paritytrading/philadelphia/acceptor/Session.java
```diff
@@ -77,7 +77,7 @@ public void reject(FIXConnection connection, FIXMessage message) throws IOExcept
             public void logon(FIXConnection connection, FIXMessage message) throws IOException {
                 connection.sendLogon(true);
 
-                connection.updateCompID(report);
+                connection.reprepare(report);
```
jheusser, line 80: This definitely reads like a typo 😆

jheusser (approved):
Looks good to me, if the overloaded versions of `prepare` wouldn't be confusing, could also set `prepare` to public 👍🏽

examples/acceptor/src/main/java/com/paritytrading/philadelphia/acceptor/Session.java
jvirtanen: I think I got the feedback I needed here. 😂

JanStureNielsen: `resetOrPrepare`? Or avoiding it with a `setOrAdd` behavior in `FIXMessage`?

jvirtanen:
Superseded by #209.

Issue #236, opened 2022-04-14
Allow to pass other channel types


ddimtirov:
In our testing, we would like to use pipe channels rather than socket channels. 

Looking at the uses of the channel, we should be able to swap the `SocketChannel` type in `FIXConnetion` with `ByteChannel`.

There would be negative impact of breaking the `SocketChannel FIXConnection::getChannel()` (it would change the return type)

Please indicate whether you would consider a Pull Request.

jvirtanen:
Hi @ddimtirov! This is an interesting idea, and I could see other use cases for this as well. However, note that [`FIXConnection::send()`](https://github.com/paritytrading/philadelphia/blob/2052dd027ef65d9f4fec37629ab09e2d88c6612e/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java#L452) uses [`ScatteringByteChannel`](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/ScatteringByteChannel.html) as well.

What do you think if we made `FIXConnection` generic with a type parameter for the concrete channel implementation?
```java
public class FIXConnection<C extends ByteChannel & ScatteringByteChannel> {
    ...
```

One downside I see is that one would need to start using the diamond syntax for a `FIXConnection` instance:
```java
FIXConnection<SocketChannel> connection = new FIXConnection<>(...);  
```

On the other hand, this would unlock pipe channels and, interestingly, even [`DatagramChannel`](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/DatagramChannel.html). 😅

ddimtirov:
Sounds good, would you like me to try it?

jvirtanen:
Please go ahead, let's see how it would look like in practice. 🙂

mrossi975:
hi guys,
the above would also allow plugging in the TLS channel implemented by the following open-source project:
https://github.com/marianobarrios/tls-channel
This is important in order to connect to CoinBase. 
Having a separate TLS tunnel in a separate tunnel is OK as a proof-of-concept but it's not great for a production environment, you have to worry about configuring, deploying and running one more process.
I am happy to help with all of the above guys..

ddimtirov:
Update - the patch is in the mail (let's see how long our internal processes would take).

mrossi975:
hey guys, any update on the above? I'd be keen to try this out. Happy to do any work required. Thanks!

jvirtanen:
This was implemented in #238.

PR #238, opened 2022-04-26
Allow to pass other channel types

ddimtirov:
Raising this pull request as a proposal for issue #236 

The change changes the raw signature of `FIXConnection.getChannel()` in backwards incompatible way. The migration fix for existing usages would be to either specify generic parameter or to downcast.

I have made the `FIXInitiatorTest` abstract and pushed the current implementation into a `FIXInitiatorSocketTest` subclass. I have also added `FIXInitiatorPipeTest`  as an example of how to use the new functionality.

### DIFF
--- a/libraries/core/src/test/java/com/paritytrading/philadelphia/FIXConnectionStatus.java
+++ b/libraries/core/src/test/java/com/paritytrading/philadelphia/FIXConnectionStatus.java
@@ -15,10 +15,12 @@
  */
 package com.paritytrading.philadelphia;
 
+import java.nio.channels.GatheringByteChannel;
+import java.nio.channels.ReadableByteChannel;
 import java.util.ArrayList;
 import java.util.List;
 
-class FIXConnectionStatus implements FIXConnectionStatusListener {
+class FIXConnectionStatus<CHANNEL extends ReadableByteChannel & GatheringByteChannel> implements FIXConnectionStatusListener<CHANNEL> {
 
     private List<Event> events;
 
@@ -31,37 +33,37 @@ class FIXConnectionStatus implements FIXConnectionStatusListener {
     }
 
     @Override
-    public void close(FIXConnection connection, String message) {
+    public void close(FIXConnection<CHANNEL> connection, String message) {
         events.add(new Close(message));
     }
 
     @Override
-    public void sequenceReset(FIXConnection connection) {
+    public void sequenceReset(FIXConnection<CHANNEL> connection) {
         events.add(new SequenceReset());
     }
 
     @Override
-    public void tooLowMsgSeqNum(FIXConnection connection, long receivedMsgSeqNum, long expectedMsgSeqNum) {
+    public void tooLowMsgSeqNum(FIXConnection<CHANNEL> connection, long receivedMsgSeqNum, long expectedMsgSeqNum) {
         events.add(new TooLowMsgSeqNum(receivedMsgSeqNum, expectedMsgSeqNum));
     }
 
     @Override
-    public void heartbeatTimeout(FIXConnection connection) {
+    public void heartbeatTimeout(FIXConnection<CHANNEL> connection) {
         events.add(new HeartbeatTimeout());
     }
 
     @Override
-    public void reject(FIXConnection connection, FIXMessage message) {
+    public void reject(FIXConnection<CHANNEL> connection, FIXMessage message) {
         events.add(new Reject());
     }
 
     @Override
-    public void logon(FIXConnection connection, FIXMessage message) {
+    public void logon(FIXConnection<CHANNEL> connection, FIXMessage message) {
         events.add(new Logon());
     }
 
     @Override
-    public void logout(FIXConnection connection, FIXMessage message) {
+    public void logout(FIXConnection<CHANNEL> connection, FIXMessage message) {
         events.add(new Logout());
     }
 
--- a/libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
+++ b/libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
@@ -24,8 +24,8 @@ import static java.util.Arrays.*;
 import static org.junit.jupiter.api.Assertions.*;
 
 import java.io.IOException;
-import java.nio.channels.ServerSocketChannel;
-import java.nio.channels.SocketChannel;
+import java.nio.channels.ByteChannel;
+import java.nio.channels.GatheringByteChannel;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import org.junit.jupiter.api.BeforeEach;
@@ -33,7 +33,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Timeout;
 
 @Timeout(value=1, unit=TimeUnit.SECONDS)
-class FIXInitiatorTest {
+abstract class FIXInitiatorTest<INITIATOR_CHANNEL extends ByteChannel & GatheringByteChannel> {
 
     private static final FIXConfig initiatorConfig = new FIXConfig.Builder()
         .setSenderCompID("initiator")
@@ -44,31 +44,28 @@ class FIXInitiatorTest {
     private FIXMessages  initiatorMessages;
     private TestMessages acceptorMessages;
 
-    private FIXConnectionStatus initiatorStatus;
+    private FIXConnectionStatus<INITIATOR_CHANNEL> initiatorStatus;
 
-    private FIXConnection  initiator;
+    private FIXConnection<INITIATOR_CHANNEL> initiator;
     private TestConnection acceptor;
+
+    final class Channels {
+        public INITIATOR_CHANNEL initiator;
+        public ByteChannel acceptor;
+    }
+
+    abstract Channels createChannels() throws IOException;
 
     @BeforeEach
     void setUp() throws IOException {
-        ServerSocketChannel acceptorServerChannel = ServerSocketChannel.open();
-        acceptorServerChannel.bind(null);
-
-        SocketChannel initiatorChannel = SocketChannel.open(acceptorServerChannel.getLocalAddress());
-        initiatorChannel.configureBlocking(false);
-
-        SocketChannel acceptorChannel = acceptorServerChannel.accept();
-        acceptorChannel.configureBlocking(false);
-
-        acceptorServerChannel.close();
-
         initiatorMessages = new FIXMessages();
         acceptorMessages  = new TestMessages();
 
-        initiatorStatus = new FIXConnectionStatus();
+        initiatorStatus = new FIXConnectionStatus<>();
 
-        initiator = new FIXConnection(initiatorChannel, initiatorConfig, initiatorMessages, initiatorStatus);
-        acceptor  = new TestConnection(acceptorChannel, acceptorMessages);
+        Channels channels = createChannels();
+        initiator = new FIXConnection<>(channels.initiator, initiatorConfig, initiatorMessages, initiatorStatus);
+        acceptor  = new TestConnection(channels.acceptor, acceptorMessages);
     }
 
     @Test
--- a/applications/client/src/main/java/com/paritytrading/philadelphia/client/Messages.java
+++ b/applications/client/src/main/java/com/paritytrading/philadelphia/client/Messages.java
@@ -19,10 +19,11 @@ import com.paritytrading.philadelphia.FIXConnection;
 import com.paritytrading.philadelphia.FIXConnectionStatusListener;
 import com.paritytrading.philadelphia.FIXMessage;
 import com.paritytrading.philadelphia.FIXMessageListener;
+import java.nio.channels.SocketChannel;
 import java.util.ArrayList;
 import java.util.List;
 
-class Messages implements FIXMessageListener, FIXConnectionStatusListener {
+class Messages implements FIXMessageListener, FIXConnectionStatusListener<SocketChannel> {
 
     private final List<Message> messages;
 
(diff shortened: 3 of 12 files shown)


mrossi975:
hi,
I have just posted a PR that is very similar to yours! Same solution using generics basically.
Great minds think alike I guess. :)
I have also posted an alternative solution that relies on a new FIXChannel interface in case Jussi doesn't like generics.
Hope it helps!

ddimtirov:
Right, I just realized that I haven't fixed all the raw usages of `FIXConnection` - will do if @jvirtanen decides to use this PR.

ddimtirov:
Fixed the raw usages of `FIXConnection` in a second commit in this PR (please squash, or let me know if I should redo with different tradeoffs).

In particular, one may argue that having precise type on the `FIXConnectionStatusListener` method parameters is not worth the added verbosity, and that we can use `FIXConnection<?>` without loss of expresivity. The latter option also allows to use the same status listener irrelevant of the underlying channel. This is important if clients are building a library of their own.

If we deprecate the `FIXConnection.getChannel()` and remove it in the 2.0 release (if anybody really wants it they can use DI to inject it again at the usage site), then we wouldn't need to generify `FIXConnectionStatusListener` and the `FIXConnection` interface would be free from generics.

jvirtanen:
Thank you, @ddimtirov! My apologies, I've been busy with work recently. I will have a look at this later this week. 🔍

ddimtirov:
Force-pushed the last commit to also fix checkstyle warnings.

jvirtanen:
@ddimtirov I just [simplified `FIXInitiatorTest`](https://github.com/paritytrading/philadelphia/commit/ee27693a6a9bf4d4fa64a36bb82426d9bf940c11) so that it no longer uses blocking I/O at all.

ddimtirov:
Rebased on top of the latest changes @jvirtanen - any ETA for when you can review the changes? Assuming you are on Europe time, we can have a quick call if you want to discuss the changes.

jvirtanen:
> Rebased on top of the latest changes.
> 
> @jvirtanen, any ETA for when you can review the changes? Assuming you are on Europe time, we can have a quick call if you want to discuss.

Great! I will have a look next week; I’m currently on holiday. 👍🏼

ddimtirov:
> Great! I will have a look next week

Hi Jussi, not to press you, but an updated ETA will be appreciated... As it would happen, this is one of the pilot cases for OSS engagement in my day-job, and I'd be keen to get it to conclusion one way or another. 

Rejection with feedback is very much an option (won't be taken personally).

jvirtanen:
> Hi Jussi, not to press you, but an updated ETA will be appreciated... As it would happen, this is one of the pilot cases for OSS engagement in my day-job, and I'd be keen to get it to conclusion one way or another.
> 
> Rejection with feedback is very much an option (won't be taken personally).

Generally, this looks good. I will have a more detailed look still today. My apologies for the delay, I currently work on this project only in my spare time.

libraries/core/src/test/java/com/paritytrading/philadelphia/DoublePipeChannel.java
```diff
@@ -0,0 +1,66 @@
+package com.paritytrading.philadelphia;
```
jvirtanen, line 1: Could you please add the license boilerplate to this file (with a 2022 copyright)?

libraries/core/src/test/java/com/paritytrading/philadelphia/DoublePipeChannel.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.ByteChannel;
+import java.nio.channels.GatheringByteChannel;
+import java.nio.channels.Pipe;
+import java.nio.channels.ScatteringByteChannel;
+
+// testing utility - in  prod code this should be selectable, and error handling more robust
+public class DoublePipeChannel implements ByteChannel, ScatteringByteChannel, GatheringByteChannel {
```
jvirtanen, line 11: Could we just call this `PipeChannel`?

I think the comment is not needed either, as this code is in the test scope in any case. To stress that further, let's make the class package-private like all other classes in this package are.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorPipeTest.java
```diff
@@ -0,0 +1,24 @@
+package com.paritytrading.philadelphia;
```
jvirtanen, line 1: Could you please add the license boilerplate to this file (with a 2022 copyright)?

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorSocketTest.java
```diff
@@ -0,0 +1,26 @@
+package com.paritytrading.philadelphia;
```
jvirtanen, line 1: Could you please add the license boilerplate to this file (with a 2022 copyright)?

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorSocketTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.io.IOException;
+import java.nio.channels.ServerSocketChannel;
+import java.nio.channels.SocketChannel;
+import org.jetbrains.annotations.NotNull;
```
jvirtanen, line 6: Let's drop this annotation, it's not been used in this project before.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
```diff
@@ -24,16 +24,17 @@
 import static org.junit.jupiter.api.Assertions.*;
 
 import java.io.IOException;
-import java.nio.channels.ServerSocketChannel;
-import java.nio.channels.SocketChannel;
+import java.nio.channels.ByteChannel;
+import java.nio.channels.GatheringByteChannel;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import org.jetbrains.annotations.NotNull;
```
jvirtanen, line 31: Let's drop this annotation, it's not been used in this project before.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorPipeTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.io.IOException;
+import java.nio.channels.Pipe;
+import org.jetbrains.annotations.NotNull;
+
+public class FIXInitiatorPipeTest extends FIXInitiatorTest<DoublePipeChannel> {
```
jvirtanen, line 7: What do you think if we called this `FIXInitiatorTestWithPipeChannel`?

Let's make also this class package-private.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorSocketTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class FIXInitiatorSocketTest extends FIXInitiatorTest<SocketChannel> {
```
jvirtanen, line 8: What do you think if we called this `FIXInitiatorTestWithSocketChannel`?

Let's make also this class package-private.

libraries/core/src/test/java/com/paritytrading/philadelphia/DoublePipeChannel.java
```diff
@@ -0,0 +1,66 @@
...
+    public long write(ByteBuffer[] srcs) throws IOException {
+        return sink.write(srcs);
+    }
+
+    @Override
+    public int write(ByteBuffer src) throws IOException {
+        return sink.write(src);
+    }
+
+
```
jvirtanen, line 40: Extra newline.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXConnectionStatus.java
```diff
@@ -15,10 +15,12 @@
  */
 package com.paritytrading.philadelphia;
 
+import java.nio.channels.GatheringByteChannel;
+import java.nio.channels.ReadableByteChannel;
 import java.util.ArrayList;
 import java.util.List;
 
-class FIXConnectionStatus implements FIXConnectionStatusListener {
+class FIXConnectionStatus<CHANNEL extends ReadableByteChannel & GatheringByteChannel> implements FIXConnectionStatusListener<CHANNEL> {
```
jvirtanen, line 23: Do we need to repeat the type parameter bounds for `CHANNEL` here?

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorPipeTest.java
```diff
@@ -0,0 +1,24 @@
+package com.paritytrading.philadelphia;
+
+import java.io.IOException;
+import java.nio.channels.Pipe;
+import org.jetbrains.annotations.NotNull;
```
jvirtanen, line 5: Let's drop this annotation, it's not been used in this project before.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
```diff
@@ -24,16 +24,17 @@
...
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Timeout;
 
 @Timeout(value=1, unit=TimeUnit.SECONDS)
-class FIXInitiatorTest {
+abstract class FIXInitiatorTest<INITIATOR_CHAN extends ByteChannel & GatheringByteChannel> {
```
jvirtanen, line 37: Could we call `INITIATOR_CHAN` `INITIATOR_CHANNEL` for consistency?

libraries/core/src/test/java/com/paritytrading/philadelphia/DoublePipeChannel.java
```diff
@@ -0,0 +1,66 @@
...
+    public DoublePipeChannel(Pipe.SinkChannel sink, Pipe.SourceChannel source) {
+        this.sink = sink;
+        this.source = source;
+    }
+
+    void configureBlocking(boolean block) throws IOException {
+        sink.configureBlocking(block);
+        source.configureBlocking(block);
+    }
```
jvirtanen, line 23: As this class is only used in non-blocking mode, perhaps the constructor could directly set both channels into non-blocking mode? Then this method wouldn't be needed at all.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
```diff
@@ -44,31 +45,29 @@ class FIXInitiatorTest {
...
-
-        SocketChannel initiatorChannel = SocketChannel.open(acceptorServerChannel.getLocalAddress());
-        initiatorChannel.configureBlocking(false);
-
-        SocketChannel acceptorChannel = acceptorServerChannel.accept();
-        acceptorChannel.configureBlocking(false);
+    protected final class Channels {
+        public INITIATOR_CHAN initiator;
+        public ByteChannel acceptor;
+    }
```
jvirtanen, line 56: This class could be in the package-private scope like other non-private members in this class.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
```diff
@@ -44,31 +45,29 @@ class FIXInitiatorTest {
...
 
-        acceptorServerChannel.close();
+    @NotNull
+    protected abstract Channels createChannels() throws IOException;
```
jvirtanen, line 59: This method could be in the package-private scope like the other non-private members in the class.

jvirtanen (changes requested):
Great work, @ddimtirov! 🙌🏼

I added a few review comments, but they are all small nits about details really. The overall design is exactly how I would have done this myself as well, including the `FIXInitiatorTest` split into a pipe channel and a socket channel variant. 👍🏼

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXConnectionStatus.java
ddimtirov: Yes we do, unless we want to hardcode it to a specific class - there are no typedefs in Java... 

On the other hand if we remove the `Session::getConnection() : FIXConnection`, we can get rid of pretty much all generics. I will raise a separate PR with that change as it will be breaking more client code and I assume it may be contentious, while the current set of changes are mostly backwards compatible.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorPipeTest.java
ddimtirov: Apologies, when IDEA sees it in the classpath, it automatically adds it :-)

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorPipeTest.java
ddimtirov: Some tools (especially those desighed in JUnit 3 timeframe) expect test classes to finish with `*Test`. The IntelliJ IDEA `Class naming convention` inspection is one of these.

Many people have similar expectations :-)

Renamed.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorSocketTest.java
ddimtirov: See my reply from the Pipe test.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
ddimtirov: Generally `protected` communicates that you intend this to be called from subclasses regardless of where they are located. This is our intent with this method and the class above - they are part of our interface to subclasses.

That said, each project has its peculiarities, and I see no harm in this.

ddimtirov (commented):
Addressed all feedback, and resolved most of the comments. 

Please have a look at the unresolved ones, as I expect you may have further feedback there.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTestWithSocketChannel.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.paritytrading.philadelphia;
+
+import java.io.IOException;
+import java.nio.channels.ServerSocketChannel;
+import java.nio.channels.SocketChannel;
+
+class FIXInitiatorTestWithSocketChannel extends FIXInitiatorTest<SocketChannel> {
```
jvirtanen, line 22: You were absolutely right about the naming conventions; JUnit 5 does not actually pick these tests up anymore on `mvn test`. 🤦🏼  Would you mind renaming these to `FIXInitiatorPipeChannelTest` and `FIXInitiatorTestSocketChannelTest` after all? Sorry! 😅

jvirtanen (approved):
Looking good! 👍🏼 

Just two more things:
1. You were right about the test class naming convention. Let's rename them once more so that they're actually run.
2. Could you please squash the commits together?

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTestWithSocketChannel.java
ddimtirov: Thanks, done.

For the record, this is a Maven and not a JUnit issue (IntelliJ and Gradle can pick these just fine). We can override the pattern of the SureFire plugin, but it is generally a bad idea with Maven.

ddimtirov:
Squashed the PR, and kept the separated commits in a different branch on my side (in case you want to separate something back).

jvirtanen:
Merged, thank you @ddimtirov! 🚀

PR #239, opened 2022-04-26
FIXConnection changes to support more Channel types (Solution 1)

mrossi975:
hi,
this solution relies on generics to allow FIXConnection to specify the 2 interfaces in Channel that it needs:

public class FIXConnection<T extends ReadableByteChannel & GatheringByteChannel> implements Closeable {

The drawbacks on this is that now FIXConnection is generic and that might not be of everyone's taste.

I will shortly post another PR with an alternative solution based on a new interface

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
@@ -23,6 +23,8 @@ import static com.paritytrading.philadelphia.FIXTags.*;
 import java.io.Closeable;
 import java.io.IOException;
 import java.nio.ByteBuffer;
+import java.nio.channels.GatheringByteChannel;
+import java.nio.channels.ReadableByteChannel;
 import java.nio.channels.SocketChannel;
 import org.joda.time.DateTimeZone;
 import org.joda.time.MutableDateTime;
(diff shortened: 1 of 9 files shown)


libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
```diff
@@ -408,18 +410,22 @@ void sendApplicationMessage() throws IOException {
         initiatorMessages(messages);
     }
```
mrossi975, line 412: hi, this is the only part where the codebase relies on blocking behaviour.
Is it still wanted / needed?

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -23,18 +23,20 @@
...
 import java.nio.ByteBuffer;
+import java.nio.channels.GatheringByteChannel;
+import java.nio.channels.ReadableByteChannel;
 import java.nio.channels.SocketChannel;
 import org.joda.time.DateTimeZone;
 import org.joda.time.MutableDateTime;
 
 /**
  * A connection.
  */
```
mrossi975, line 34: this is basically the root of the proposed solution

jvirtanen:
Thank you, @mrossi975! My apologies, I've been busy with work recently. I will have a look at these later this week. 🔍

jvirtanen:
Thank you for this contribution, @mrossi975! 🙌🏼  Support for this feature just got merged in #238.

PR #250, opened 2022-06-21
Clean up the API by removing FIXConnection.getChannel()

ddimtirov:
Removing `getChannel()` and introducing a second connection constructor with upstream and downstream pipe, allows us to get rid of virtually all generics and simplify the pipe code. It is also increases the encapsulation of the connection opening the possibility of alternative (non-channel based) transports for exotic scenarios, without affecting client code.

While the `getChannel()` may look like logical necessity, in practice it was not used in this codebase and I don't find usages in my codebase. If I wanted to get a handle of it, I could have gotten it at the place of constructing the connection.


libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
```diff
@@ -61,10 +63,13 @@ void setUp() throws IOException {
         initiatorMessages = new FIXMessages();
         acceptorMessages  = new TestMessages();
 
-        initiatorStatus = new FIXConnectionStatus<>();
+        initiatorStatus = new FIXConnectionStatus();
 
         Channels channels = createChannels();
-        initiator = new FIXConnection<>(channels.initiator, initiatorConfig, initiatorMessages, initiatorStatus);
+        initiator = new FIXConnection(
+                channels.initiatorRead, channels.initiatorWrite,
```
ddimtirov, line 70: With this change, separating the roles, we won't need the equivalent of out `PipeChannel` testing class if we want to use pipes in PROD code - we can use the upstream and downstream pipes directly.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorPipeChannelTest.java
```diff
@@ -18,14 +18,15 @@
...
+class FIXInitiatorPipeChannelTest extends FIXInitiatorTest {
     @Override
     protected Channels createChannels() throws IOException {
         Pipe up = Pipe.open();
         Pipe down = Pipe.open();
 
         Channels channels = new Channels();
-        channels.initiator = new PipeChannel(up.sink(), down.source());
+        channels.initiatorRead = down.source();
+        channels.initiatorWrite = up.sink();
```
ddimtirov, line 29: Note that we are not using the `PipeChannel` for the `FIXConnection`, and we could have rewritten the acceptor class to also not need it (if we segregate the upstream and downstream channels).

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -29,11 +29,12 @@
...
  * A connection.
  */
-public class FIXConnection<CHANNEL extends ReadableByteChannel & GatheringByteChannel> implements Closeable {
+public class FIXConnection implements Closeable {
 
     private static final int CURRENT_TIMESTAMP_FIELD_CAPACITY = 24;
 
-    private final CHANNEL channel;
+    private final ReadableByteChannel readableChannel;
+    private final GatheringByteChannel writableChannel;
```
ddimtirov, line 37: We are segregating the upstream and downstream channels - this serves two purposes - work around Java not having union types, and simplify cases like pipe transport, where channels are unidirectional.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -92,14 +93,30 @@
...
      * @param listener the inbound message listener
      * @param statusListener the inbound status event listener
      */
-    public FIXConnection(CHANNEL channel, FIXConfig config, FIXMessageListener listener,
-            FIXConnectionStatusListener<CHANNEL> statusListener) {
-        this.channel = channel;
+    public <CHANNEL extends ReadableByteChannel & GatheringByteChannel>
+    FIXConnection(CHANNEL channel, FIXConfig config, FIXMessageListener listener, FIXConnectionStatusListener statusListener) {
+            this(channel, channel, config, listener, statusListener);
+    }
```
ddimtirov, line 104: We preserve the old constructor, so this change should not break client code.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -153,15 +170,6 @@ public FIXConnection(CHANNEL channel, FIXConfig config, FIXMessageListener liste
         this.currentTimestamp.setTimestampMillis(this.currentTime);
     }
 
-    /**
-     * Get the underlying socket channel.
-     *
-     * @return the underlying socket channel
-     */
-    public CHANNEL getChannel() {
```
ddimtirov, line 161: This change will break client code, but is easy to work around by passing the channel by other means (DI, static, threadlocal are plausible workarounds work if one needs to upgrade, but cannot afford to refactor right now).

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -387,7 +395,8 @@ public void keepAlive() throws IOException {
      */
     @Override
     public void close() throws IOException {
-        channel.close();
+        readableChannel.close();
+        writableChannel.close();
     }
```
ddimtirov, line 400: These are iffy - ideally the second close should have been in finally, or should have caught the 2 exceptions and thrown another one with those added as suppressed.

ddimtirov (commented):
Added some comments in the interesting spots - most/all of the changes in the other files are removing generics and imports.

ddimtirov:
@jvirtanen please and let me know if you approve of the idea - if yes, I'll fix the failing checks.

If no - please close :-)

jvirtanen:
Hi @ddimtirov! Sorry, I missed this earlier in the week; I will have a look. 🔍

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorPipeChannelTest.java
jvirtanen: If we can get rid of `PipeChannel` as an abstraction, then I think it would be worthwhile to do the change also on the acceptor side.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
jvirtanen: Yes, let's wrap this so that both methods get called even if the first throws an exception. Maybe something like this could work? 🤔 
```java
IOException readableException = null;
IOException writableException = null;

try {
    readableChannel.close();
} catch (IOException e) {
    readableException = e;
}

try {
    writableChannel.close();
} catch (IOException e) {
    writableException = e;
}

if (writableException != null) {
    if (readableException != null)
        writableException.addSuppressed(readableException);

    throw writableException;
}
else if (readableException != null) {
    throw readableException;
}
```

jvirtanen (changes requested):
Nice work, @ddimtirov! 🙌🏼 

I feel that the challenges in adopting Philadelphia to work with other transports than `SocketChannel` strengthen how I'm feeling this library should ultimately work: just provide methods to parse and format complete FIX messages from and to various buffers, leaving the reading and writing of these buffers outside the scope of the library. This would also improve interoperability with libraries such as Aeron and Netty.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
ddimtirov: My personal view is that I'd rather go without. This improves the encapsulation of the class, and if someone wants access to the channels there are many easy ways of doing it (i.e. the class that created the channels is a better place to look them up from).

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
ddimtirov: done

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
ddimtirov: How about

```
        try (Closeable ignored1 = rxChannel; Closeable ignored2 = txChannel) {
            // we just want to close both channels reliably  
        }
```

ddimtirov:
Addressed all feedback, please have a look at the `close()` method - it looks weird, but it is compact and clean.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -386,8 +395,11 @@ public void keepAlive() throws IOException {
      * @throws IOException if an I/O error occurs
      */
     @Override
+    @SuppressWarnings("EmptyTryBlock")
```
jvirtanen, line 398: I think this is IntelliJ IDEA'ism? I would rather leave it out. 🙂

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
jvirtanen: Alright.

libraries/core/src/test/java/com/paritytrading/philadelphia/TestConnection.java
```diff
@@ -70,12 +72,15 @@ void send(List<String> messages) throws IOException {
         txBuffer.flip();
 
         while (txBuffer.hasRemaining())
-            channel.write(txBuffer);
+            txChannel.write(txBuffer);
     }
 
     @Override
+    @SuppressWarnings("EmptyTryBlock")
```
jvirtanen, line 79: See the comment in `FIXConnection`.

libraries/core/src/test/java/com/paritytrading/philadelphia/TestConnection.java
```diff
@@ -70,12 +72,15 @@ void send(List<String> messages) throws IOException {
...
     public void close() throws IOException {
-        channel.close();
+        try (Closeable ignored1 = rxChannel; Closeable ignored2 = txChannel) {
+            // we just want to close both channels reliably
+        }
```
jvirtanen, line 83: See the comment in `FIXConnection`.

jvirtanen (changes requested):
Looking good, @ddimtirov! 🙌🏼 I just left a few more comments, and then, as last time, could you please squash the changes into a single commit? 🙂

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
ddimtirov: Getting rid of the ones I added. There is one more in `FIXConfigTest` (and yes, IntelliJ adds these, but it is a standard Java mechanism to suppress a warning. I.e. `javac` also uses it.)

ddimtirov:
Done, the split commits (easier to review) are available in https://github.com/ddimtirov/philadelphia/tree/philadelphia-250-maintenance

jvirtanen (approved):

jvirtanen:
Thank you, @ddimtirov! 🚀

ddimtirov:
@jvirtanen, what about a release? Is there anything that you would like done before that?

jvirtanen:
@ddimtirov Yes! I've started some preparations for the 2.0.0 release. 👍🏼

ddimtirov:
Let me know if you need any help (tests, measurements, doc drafts, etc.)

PR #280, opened 2022-08-04
Improve 'FIXMessageListener'

jvirtanen:
Add a `connection` parameter to the `message` method to align the class with the `FIXConnectionStatusListener` interface. This is helpful for separating the administrative message handling from `FIXConnection` as a `FIXMessageListener` that invokes the `FIXConnectionStatusListener` methods.


jvirtanen:
@ddimtirov I would appreciate a comment from you as you've been working on the most recent major API changes heading to the Philadelphia 2.0.0 release.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessageParser.java
```diff
@@ -25,6 +25,8 @@
  */
 public class FIXMessageParser {
 
+    private final FIXConnection connection;
```
ddimtirov, line 28: We are keeping the connection in a field, without actually using it, only to be able to pass it through. To me that is a design smell. If this class was inlined in `FIXConnection` we would be able to pass `this`.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessageParser.java
```diff
@@ -25,6 +25,8 @@
  */
 public class FIXMessageParser {
```
ddimtirov, line 26: Considering that we cannot customize the message parser (there is no constructor argument and it is `new`-ed in the constructor), it would be better off if this class was inlined in `FIXConnection`.

Not a hard opinion.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -133,7 +133,7 @@ public FIXConnection(ReadableByteChannel inboundChannel, GatheringByteChannel ou
         this.senderCompId = config.getSenderCompID();
         this.targetCompId = config.getTargetCompID();
 
-        this.parser = new FIXMessageParser(config, new MessageHandler(listener));
+        this.parser = new FIXMessageParser(config, this, new MessageHandler(listener));
```
ddimtirov, line 136: `MessageHandler` being a private inner class is an issue. In particular, at some point, I needed to customize some of the handling and I could not subclass and override. I used ByteBuddy to intercept the method, and later removed the feature as I found a way ton achieve the same in different part of the application.

My view is that we should inline the message parser in the connection class (using protected methods, so people can customize), and pass the message handler from outside, providing default implementation on `null`.

ddimtirov (approved):
I am not sure I like the bidirectional coupling of the connection and the listener. Right now, if I want to achieve this, I would rather inject the connection into a listener with `init()` method (two phase initialization). 

Another way is to change the connection, so that user can set the listener after creation. If we are willing to have an array in the inner loop, we may even support multiple listeners (convenient for debugging).

Alternatively, if we consider a refactoring of the `FIXConnection` `FIXMessageParser` and `FIXConnection$MessageHAndler` classes, we may achieve better factoring of the responsibilities and make one of the above options fit more naturally.

Apologies for the rambling review, but as I was reading this code it seemed worth mentioning. Let me know if you want a strawman PR of what such a refactoring may look like.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXMessageParser.java
jvirtanen: I agree with the design smell.

`FIXMessageParser` has been a separate class from the beginning, not least for the reason that it is possible to write [a microbenchmark](https://github.com/paritytrading/philadelphia/blob/main/tests/perf-test/src/main/java/com/paritytrading/philadelphia/FIXMessageParserBenchmark.java) for it. In [1.1.0](https://github.com/paritytrading/philadelphia/releases/1.1.0), it became part of the public API (#51, #83). Retaining this use case would be nice.

One option would be to retain `FIXMessageListener` and `FIXMessageParser` as they are and to introduce `FIXConnectionListener` that passes a `connection` parameter to the `message` method. Then `FIXConnection` would bridge `FIXMessageListener` into `FIXConnectionListener`. It would avoid this particular design smell.

What do you think?

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
jvirtanen: I agree on the `MessageHandler` issue. The plan is to move it out from `FIXConnection` and make it possible for the users of the library to provide their own replacement if they wish. Currently `MessageHandler` implements a subset of the FIX protocol, but this subset is currently too little for some applications and unnecessarily much for others.

PR #302, opened 2022-09-10
Proper initialization of lastTxMillis and lastRxMillis.

### LINKED ISSUE #301: Proper initialization of lastTxMillis and lastRxMillis required
`lastTxMillis` and `lastRxMillis` in [FIXConnection](https://github.com/paritytrading/philadelphia/blob/765c1c5123d761aa411c32e03e7d186ccdec98fc/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java) are both initialized with 0.

If `Keepalive()` is called before any data reception or transmission, e.g. in the [Receiver of the sample application](https://github.com/paritytrading/philadelphia/blob/765c1c5123d761aa411c32e03e7d186ccdec98fc/applications/client/src/main/java/com/paritytrading/philadelphia/client/Session.java), `Keepalive()` triggers a `sendHeartbeat()` and a `sendTestRequest()`.

No other message than a Logon message should be the first message after a connect.

My quickfix is an initialization of `lastTxMillis` and `lastRxMillis` with `System.currentTimeMillis()` so `sendHeartbeat()` and a `sendTestRequest()` are not immediatly triggered after a connect.

dmech:
Current initialization with 0 leads to potential messages before the Logon message.

Fixes #301

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
@@ -100,13 +100,14 @@ public class FIXConnection implements Closeable {
      * @param config the connection configuration
      * @param listener the inbound message listener
      * @param statusListener the inbound status event listener
+     * @param currentTimeMillis the current time in milliseconds
      * @param <CHANNEL> the underlying channel type, which must implement
      *   {@link ReadableByteChannel} and {@link GatheringByteChannel}
      * @see SocketChannel
      */
     public <CHANNEL extends ReadableByteChannel & GatheringByteChannel>
-    FIXConnection(CHANNEL channel, FIXConfig config, FIXMessageListener listener, FIXConnectionStatusListener statusListener) {
-        this(channel, channel, config, listener, statusListener);
+    FIXConnection(CHANNEL channel, FIXConfig config, FIXMessageListener listener, FIXConnectionStatusListener statusListener, long currentTimeMillis) {
+        this(channel, channel, config, listener, statusListener, currentTimeMillis);
     }
 
     /**
@@ -119,9 +120,10 @@ public class FIXConnection implements Closeable {
      * @param config the connection configuration
      * @param listener the inbound message listener
      * @param statusListener the inbound status event listener
+     * @param currentTimeMillis the current time in milliseconds
      */
     public FIXConnection(ReadableByteChannel inboundChannel, GatheringByteChannel outboundChannel,
-                         FIXConfig config, FIXMessageListener listener, FIXConnectionStatusListener statusListener) {
+                         FIXConfig config, FIXMessageListener listener, FIXConnectionStatusListener statusListener, long currentTimeMillis) {
         this.rxChannel = inboundChannel;
         this.txChannel = outboundChannel;
 
@@ -158,8 +160,8 @@ public class FIXConnection implements Closeable {
         this.txBuffers[0] = txHeaderBuffer;
         this.txBuffers[1] = txBodyBuffer;
 
-        this.lastRxMillis = 0;
-        this.lastTxMillis = 0;
+        this.lastRxMillis = currentTimeMillis;
+        this.lastTxMillis = currentTimeMillis;
 
         this.heartbeatMillis = config.getHeartBtInt() * 1000;
 
@@ -169,9 +171,10 @@ public class FIXConnection implements Closeable {
 
         this.adminMessage = new FIXMessage(ADMIN_MESSAGE_MAX_FIELD_COUNT, config.getFieldCapacity());
 
-        this.currentTimeMillis = 0;
+        this.currentTimeMillis = currentTimeMillis;
 
         this.currentTime = new FIXTimestamp();
+        this.currentTime.setEpochMilli(currentTimeMillis);
 
         this.currentTimestamp = new FIXValue(CURRENT_TIMESTAMP_FIELD_CAPACITY);
 
--- a/libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
+++ b/libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
@@ -68,7 +68,7 @@ abstract class FIXInitiatorTest {
         Channels channels = createChannels();
         initiator = new FIXConnection(
                 channels.initiatorRx, channels.initiatorTx,
-                initiatorConfig, initiatorMessages, initiatorStatus
+                initiatorConfig, initiatorMessages, initiatorStatus, 0
         );
         acceptor  = new TestConnection(channels.acceptorRx, channels.acceptorTx, acceptorMessages);
     }
--- a/applications/client/src/main/java/com/paritytrading/philadelphia/client/Session.java
+++ b/applications/client/src/main/java/com/paritytrading/philadelphia/client/Session.java
@@ -68,7 +68,7 @@ class Session implements Closeable {
 
         channel.register(selector, SelectionKey.OP_READ);
 
-        FIXConnection connection = new FIXConnection(channel, config, listener, statusListener);
+        FIXConnection connection = new FIXConnection(channel, config, listener, statusListener, System.currentTimeMillis());
 
         return new Session(selector, connection);
     }
--- a/examples/acceptor/src/main/java/com/paritytrading/philadelphia/acceptor/Session.java
+++ b/examples/acceptor/src/main/java/com/paritytrading/philadelphia/acceptor/Session.java
@@ -84,7 +84,7 @@ class Session implements FIXMessageListener {
                 connection.sendLogout();
             }
 
-        });
+        }, System.currentTimeMillis());
 
         report = connection.create();
 
--- a/examples/initiator/src/main/java/com/paritytrading/philadelphia/initiator/Initiator.java
+++ b/examples/initiator/src/main/java/com/paritytrading/philadelphia/initiator/Initiator.java
@@ -79,7 +79,7 @@ class Initiator implements FIXMessageListener, Closeable {
                 connection.sendLogout();
             }
 
-        });
+        }, System.currentTimeMillis());
 
         histogram = new Histogram(3);
     }
--- a/libraries/core/src/test/java/com/paritytrading/philadelphia/FIXConfigTest.java
+++ b/libraries/core/src/test/java/com/paritytrading/philadelphia/FIXConfigTest.java
@@ -27,7 +27,7 @@ class FIXConfigTest {
         FIXConfig config = FIXConfig.newBuilder()
                 .setVersion(FIXVersion.FIXT_1_1)
                 .build();
-        new FIXConnection(null, config, null, null);
+        new FIXConnection(null, config, null, null, 0);
     }
 
     @Test


dmech:
My first pull request and  first time Junit. Inputs are welcome.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
```diff
@@ -399,6 +404,21 @@ void receiveTooLongMessage() throws IOException {
...
+    	initiator.keepAlive();
+    	
+    	initiator.sendLogon(false);
+    	
+    	acceptor.receive();
+    	
+    	char firstMessageType = FIXMessage.fromString(acceptorMessages.collect().get(0)).getMsgType().asChar();
+    	
+        assertEquals(FIXMsgTypes.Logon, firstMessageType);        
+    }
```
jvirtanen, line 421: Could we make this test work so that it would rather just check that if `initiator.keepAlive()` is invoked _before_ `initiator.setCurrentTimeMillis()` is invoked, `initiatorStatus.collect()` would still return an empty list? See e.g. the `heartbeatTimeout` test for reference.

As I understand, the above would have failed before the fix and will now ensure that the issue cannot be accidentally reintroduced.

jvirtanen (changes requested):
Thank you for the contribution, @dmech! See below for a suggestion that should simplify the test and possibly the implementation under test as well (as a new public method is no longer needed).

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
dmech: Thank you for your input @jvirtanen.

### Public method:
I thought about the new public method as well, because I don't like to make code worse, just to have a running testcase.

Having the `currentTimestamp` set during the construction of the `FixConnection` leads to a different response in the field 52 than "19700101-00:00:00.000".
Many testcases rely on this "19700101-00:00:00.000", this is why I introduced the new public method, to be able to reset `currentTimestamp` to 0 for the tests.
An exception is the new test `firstMessageAfterConnect()` , which requires a proper time, otherwise, we would have no change compared to the old constructor.

--> The public Method is only removable, if we modify all tests with field 52 to use the current time instead of "19700101-00:00:00.000". This could make the new testcase  `firstMessageAfterConnect()` obsolete and also the simplification below.



### About your suggestion to simplify the testcase, I like your idea to have this:
```
    	initiator.keepAlive();
    	initiator.setCurrentTimeMillis(System.currentTimeMillis());
    	
    	assertEquals(asList(), initiatorStatus.collect());
```

But, sending the test/heartbeat messages in `keepAlive()` will not add messages to initiatorStatus. Only the event "heartbeatTimeout" does, which is called 30 seconds after the first timeout.
So `initiatorStatus` will be empty, equal if the time-variables in constructor are initialized by 0 or by the current time.

You are correct, the testcase is to prevent an accidental reintroduction of the issue.

--> We could modify the `send()` method in `TestConnection` to have access to all sent messages, or keep the test as in the pull request, which I agree isn't a beauty.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -169,12 +166,27 @@ public FIXConnection(ReadableByteChannel inboundChannel, GatheringByteChannel ou
 
         this.adminMessage = new FIXMessage(ADMIN_MESSAGE_MAX_FIELD_COUNT, config.getFieldCapacity());
 
-        this.currentTimeMillis = 0;
-
         this.currentTime = new FIXTimestamp();
 
         this.currentTimestamp = new FIXValue(CURRENT_TIMESTAMP_FIELD_CAPACITY);
+        
+        initializeTime(System.currentTimeMillis());
```
ddimtirov, line 173: This is a pet peeve of mine - using system time directly in a library class can lead to non-reproducible tests.

In this case there is an easy workaround - call `initializeTime()` before you send the first message, but then the design allows you not to do it too.

One possible way is to add a constructor parameter. 

Another is to set the times to -1 and throw informative exception if time is not initialized properly when sending a message. This will make it slightly more cumbersome to use, but will achieve the following: 

1. time will be initialized with user-controlled value, making it more test friendly
2. would remind people who don't read docs that handling time is their responsibility (I admit I missed that the first time around)

dmech: I agree, adding a constructor is a good and test friendly way to get rid of the public method.
As we are talking about a new major version, it is also legit to change some constructor parameters.

Regarding the informative exception I don't like checking proper initialization in a core function for performance reasons, even if we are talking about micro-optimizations. It would also require to keep the public `initializeTime()` method.

ddimtirov: The check can be an assertion, so those concerned about the last drop of performance can disable it (though they should be using FPGA by now). Pretty much all build tools run tests with assertions enabled, so that should be good enough.

libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
jvirtanen: Good catch regarding the test case. We should check that `initiatorMessages` rather than `initiatorStatus` is empty here. See e.g. the [`heartbeat`](https://github.com/paritytrading/philadelphia/blob/main/libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java#L77) test case for an example.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
dmech: I implemented the new constructor [here](https://github.com/paritytrading/philadelphia/pull/302/commits/92f31939b3da519b57026813d15f0bf9f42ac043).
As written in the discussion about the testcase, the testcase is omited, as I think, one will think twice, before removing a parameter in the constructor of a core class. To reach the main goal of the test, which is to prevent the reintroduction of the inproper initialization of the time variables, the new parameter should do the same job.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -100,13 +100,15 @@ public class FIXConnection implements Closeable {
      * @param config the connection configuration
      * @param listener the inbound message listener
      * @param statusListener the inbound status event listener
+     * @param initialMillis the initial time for the connection in epoch time.
+     *   For production normally System.currentTimeMillis()
```
jvirtanen, line 104: Let's align this with [`setCurrentTimeMillis()`](https://github.com/paritytrading/philadelphia/blob/41051aad4679ca3823b342db059e40f7bc582c41/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java#L336): `currentTimeMillis`, "the current time in milliseconds"

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -169,12 +170,18 @@ public FIXConnection(ReadableByteChannel inboundChannel, GatheringByteChannel ou
...
-
+        
+        this.lastRxMillis = initialMillis;
+        
+        this.lastTxMillis = initialMillis;
+        
+    	this.currentTimeMillis = initialMillis;
+    	
+    	this.currentTime.setEpochMilli(initialMillis);
+
```
jvirtanen, line 184: Let's minimize the diff in this method.

jvirtanen (changes requested):
Looking good! 🙌🏼 See below for a few more nits about formatting.

Also, could you please squash all the commits into a single one at the end?

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
dmech: alignment done in [29ebc04](https://github.com/paritytrading/philadelphia/pull/302/commits/29ebc049f7e659daabc7aca7264b2a1bf0820b9d)

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
dmech: minimized in [29ebc04](https://github.com/paritytrading/philadelphia/pull/302/commits/29ebc049f7e659daabc7aca7264b2a1bf0820b9d)

dmech:
> Looking good! 🙌🏼 See below for a few more nits about formatting.
> 
> Also, could you please squash all the commits into a single one at the end?

Thanks!
I hope all requirements, including the squash, are filled with [29ebc04](https://github.com/paritytrading/philadelphia/pull/302/commits/29ebc049f7e659daabc7aca7264b2a1bf0820b9d)

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -119,9 +121,11 @@ public class FIXConnection implements Closeable {
      * @param config the connection configuration
      * @param listener the inbound message listener
      * @param statusListener the inbound status event listener
+     * @param initialMillis the current time in milliseconds
+     *   For production normally System.currentTimeMillis()
```
jvirtanen, line 125: See above.

jvirtanen (changes requested):
Please see below for one more nit regarding the Javadoc documentation. 🤓 

Also, could you please set your Git author information correctly; now the commit is attributed to "jvirtanen \<test@test.com>". 😅

jvirtanen:
Now the commit is attributed to me. Can you please put in your own name and email address instead? 🙂

dmech:
> Now the commit is attributed to me. Can you please put in your own name and email address instead? slightly_smiling_face

At least I get used with the squash process :muscle: :nerd_face:

Sorry for the mess with the author and the committer. In eclipse everything seems to be fine :frowning_face:
So, if I understand, the author is correct now, and you want me to put my full name instead my username into the committer?

jvirtanen:
> Sorry for the mess with the author and the committer. In eclipse everything seems to be fine ☹️ So, if I understand, the author is correct now, and you want me to put my full name instead my username into the committer?

No worries. 🙂  If you have access to Git on the command line, you can change the author information for the commit as follows:
```
git commit --amend --author "Alice Example <alice@example.com>"
```
The commit author can either be your full name or a pseudonym if you prefer, but the associated email address should be valid. If you want to use a pseudonym, one option is to set the email address to the one generated by GitHub. You can see this in the [Email settings](https://github.com/settings/emails) under the *Keep my email addresses private* heading; it looks probably something like 123456+dmech@users.noreply.github.com for you.

dmech:
Thanks! I changed the committer to my pseudonym and the github email.
I left the author on you, as you are the original author and it feels wrong to change it to me.
Maybe I misunderstood the meaning of author and the author is related to the author of the commit and not to the author of the original file. But in this case, what is the idea of the committer attribute..?

ddimtirov:
Hi dmech, you should set yourself as both committer and author in your branch. When Jussi merges the change he becomes the committer, and if I cherrypick the change in my private branch, I will become the committer, while you will remain the author in all commits.

From the git docs:

> You may be wondering what the difference is between author and committer. The author is the person who originally wrote the work, whereas the committer is the person who last applied the work. So, if you send in a patch to a project and one of the core members applies the patch, both of you get credit — you as the author, and the core member as the committer. We’ll cover this distinction a bit more in Distributed Git.

https://git-scm.com/book/ms/v2/Git-Basics-Viewing-the-Commit-History

dmech:
Thank you ddimtirov!
Now it makes sense for me. I didn't think about the merge which will modify the attributes again.

jvirtanen (approved):
Thank you, @dmech! 🙌🏼

Issue #315, opened 2022-09-27
handleLogon() not triggered in case of unexpected msgSeqNum


dmech:
According to the FIX-specification, chapter "Synchronization after successful logon" --> [Link](https://www.fixtrading.org/standards/fix-session-layer-online/) the acknowledgment for the logon is transferred, before the resend request for possible missing message is sent.
Philadelphia handles this well, except the logon message is not triggered in the case when `msgSeqNum != rxMsgSeqNum`

The sequence number comparison in the MessageHandler of FIXConnection could/should be extended with for example:
```
if (msgSeqNum != rxMsgSeqNum) {
	if (msgType.byteAt(0) == Logon) {
		handleLogon(message);
	}
	handleMsgSeqNum(message, msgType, msgSeqNum);
	return;
}
```

jvirtanen:
Good catch, @dmech! 🔍 

As I read the section, it looks though that this would only apply if `msgSeqNum > rxMsgSeqNum`, right? Nevertheless, we should fix the handling for that case.

I suggest adding the handling inside `MessageHandler#handleMsgSeqNum()` in order to keep it off the fast path of `MessageHandler#message()`, which we try to keep as simple as possible.

dmech:
I rechecked the section and you are right, in this section the case msgSeqNum < rxMsgSeqNum is not mentioned, but in one section above it is:

> If MsgSeqNum(34) is present in the incoming Logon(35=A) message and is less than NextNumIn, then a Logout(35=5) message should be sent assuming that an error exists in the state of either the initiator or acceptor FIX session processor, followed by a termination of the transport layer connection.


At the moment in this case (msgSeqNum < rxMsgSeqNum) a `handleTooLowMsgSeqNum()` is called. Looking at the quote above, for a Logon message, a `handleLogout()` should be called in `MessageHandler#handleTooLowMsgSeqNum()` not only for the Logoff case, but also for the Logon case.

Handling in `MessageHandler#handleMsgSeqNum()` is a good idea.
For the implementation I found different approaches to identify a messagetype from a `FIXValue`:
- once in `MessageHandler#message()` with .byteAt(0) == `FIXMessageType`
- once in `MessageHandler#handleTooLowMsgSeqNum()` with `.contentEquals(FIXMessageType)`

which is the preferred way?

jvirtanen:
The preferred way outside `MessageHandler#message()` is to use `FIXValue#contentEquals()`, which is more descriptive.

[...]

PR #320, opened 2022-10-09
Improve handling of logon acknowledgements if msgSeqNum != rxMsgSeqNum

dmech:
See #315

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
@@ -598,6 +598,21 @@ public class FIXConnection implements Closeable {
         send(adminMessage);
     }
 
+    /**
+     * Send a ResendRequest(2) message.
+     *
+     * @param beginSeqNo the BeginSeqNo(7)
+     * @throws IOException if an I/O error occurs
+     */
+    public void sendResendRequest(long beginSeqNo) throws IOException {
+        prepare(adminMessage, ResendRequest);
+
+        adminMessage.addField(BeginSeqNo).setInt(beginSeqNo);
+        adminMessage.addField(EndSeqNo).setInt(0);
+
+        send(adminMessage);
+    }
+
     private void sendHeartbeat() throws IOException {
         prepare(adminMessage, Heartbeat);
 
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnectionStatusHandler.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnectionStatusHandler.java
@@ -100,16 +100,23 @@ class FIXConnectionStatusHandler implements FIXMessageListener {
     private void handleMsgSeqNum(FIXMessage message, FIXValue msgType, long msgSeqNum) throws IOException {
         long rxMsgSeqNum = connection.getInMsgSeqNum();
 
-        if (msgSeqNum < rxMsgSeqNum)
+        if (msgSeqNum < rxMsgSeqNum) {
             handleTooLowMsgSeqNum(message, msgType, msgSeqNum);
-        else
-            sendResendRequest(rxMsgSeqNum);
+        } else {
+            if (msgType.contentEquals(Logon))
+                handleLogon(message);
+            else
+                connection.sendResendRequest(rxMsgSeqNum);
+        }
     }
 
     private void handleTooLowMsgSeqNum(FIXMessage message, FIXValue msgType, long msgSeqNum) throws IOException {
         long rxMsgSeqNum = connection.getInMsgSeqNum();
 
-        if (msgType.contentEquals(Logout)) {
+        if (msgType.contentEquals(Logon)) {
+            connection.sendLogout("MsgSeqNum(34) too low");
+            statusListener.close(connection, "MsgSeqNum(34) too low");
+        } else if (msgType.contentEquals(Logout)) {
             handleLogout(message);
         } else if (!msgType.contentEquals(SequenceReset)) {
             FIXValue possDupFlag = message.valueOf(PossDupFlag);
@@ -221,15 +228,6 @@ class FIXConnectionStatusHandler implements FIXMessageListener {
         connection.prepare(adminMessage, Heartbeat);
 
         adminMessage.addField(TestReqID).set(testReqId);
-
-        connection.send(adminMessage);
-    }
-
-    private void sendResendRequest(long beginSeqNo) throws IOException {
-        connection.prepare(adminMessage, ResendRequest);
-
-        adminMessage.addField(BeginSeqNo).setInt(beginSeqNo);
-        adminMessage.addField(EndSeqNo).setInt(0);
 
         connection.send(adminMessage);
     }
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnectionStatusListener.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnectionStatusListener.java
@@ -62,9 +62,15 @@ public interface FIXConnectionStatusListener {
     /**
      * Receive a Logon(A) message.
      *
+     * <p>Please note that the incoming Logon(A) message might have a MsgSeqNum(34)
+     * value that is higher than the next expected MsgSeqNum(34) value.
+     * In this case, the application should immediately follow the Logon(A)
+     * response with a ResendRequest(2) message.</p>
+     *
      * @param connection the connection
      * @param message the Logon(A) message
      * @throws IOException if an I/O error occurs
+     * @see FIXConnection#sendResendRequest(long beginSeqNo)
      */
     void logon(FIXConnection connection, FIXMessage message) throws IOException;
 
--- a/libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
+++ b/libraries/core/src/test/java/com/paritytrading/philadelphia/FIXInitiatorTest.java
@@ -348,6 +348,28 @@ abstract class FIXInitiatorTest {
     }
 
     @Test
+    void receiveLogonWithTooLowMsgSeqNum() throws IOException {
+        initiator.setInMsgSeqNum(2);
+
+        String message = "35=A|34=1|";
+        Event  status  = new Close("MsgSeqNum(34) too low");
+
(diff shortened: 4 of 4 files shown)


libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -605,12 +605,16 @@ public void message(FIXMessage message) throws IOException {
...
+                if (msgType.contentEquals(Logon)) {
+                    handleLogon(message);
+                }
+            }
         }
 
         private void handleTooLowMsgSeqNum(FIXMessage message, FIXValue msgType, long msgSeqNum) throws IOException {
-            if (msgType.contentEquals(Logout)) {
+            if (msgType.contentEquals(Logout) || msgType.contentEquals(Logon)) {
                 handleLogout(message);
```
jvirtanen, line 618: Should this perhaps look as follows?
```java
if (msgType.contentEquals(Logon)) {
    handleLogon(message);
} else if (msgType.contentEquals(Logout)) {
    handleLogout(message);
…
```

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -605,12 +605,16 @@ public void message(FIXMessage message) throws IOException {
         private void handleMsgSeqNum(FIXMessage message, FIXValue msgType, long msgSeqNum) throws IOException {
             if (msgSeqNum < rxMsgSeqNum)
                 handleTooLowMsgSeqNum(message, msgType, msgSeqNum);
-            else
+            else {
                 sendResendRequest(rxMsgSeqNum);
+                if (msgType.contentEquals(Logon)) {
+                    handleLogon(message);
+                }
```
jvirtanen, line 612: Should this block be before or after the `sendResendRequest()` invocation? Typically `handleLogon()` would respond with a Logon(A) message. Therefore we would now end up sending ResendRequest(2) first and only then Logon(A). However, there’s still the possibility that `handleLogon()` would kick off an asynchronous operation. In that case, even reordering these two methods would not be enough to ensure that Logon(A) is sent out first. 🤔

jvirtanen (commented):
Thank you, @dmech! See below for a few comments. In addition, I would like to see test cases for the new functionality in `FIXInitiatorTest`. 🙂

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
dmech: Thank you for this suggestion @jvirtanen 

My idea is based on the FIX specs quoted here:  [Issuecomment](https://github.com/paritytrading/philadelphia/issues/315#issuecomment-1264348589)
But you are right, the Logon case should lead to a different call as the Logout case does, although both cases are Log**out** related.
How about this:

```java
if (msgType.contentEquals(Logon)) {
    sendLogout();
    statusListener.close(FIXConnection.this, "MsgSeqNum(34) too low");
} else if (msgType.contentEquals(Logout)) {
    handleLogout(message);
…

```

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
dmech: I agree, the reordered way is the cleaner one. Until now I was looking at the Issue from the initiator side and did not think about the acceptor side.
In my application I have such an asynchronous operation. And I handle it with a 1 second sleep before the operation. For sure a hacky workaround, but for the moment I had no better idea.

dmech:
Thank you @jvirtanen :+1: , I will add some tests

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
jvirtanen: Right, good point. I think the above would be a good approach; we already send Logout if the MsgSeqNum(34) field is missing. However, perhaps we could set the Text(58) field in the Logout(5) message also in this case:
```java
sendLogout("MsgSeqNum(34) too low");
```

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
jvirtanen: I thought a bit about this.

I think what we should do here in case of a Logon(A) message is to invoke `handleLogon()` and _not_ invoke `sendResendRequest()`. Then in the documentation of [`FIXConnectionStatusListener#logon()`](https://github.com/paritytrading/philadelphia/blob/main/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnectionStatusListener.java#L69) we should note that the incoming Logon(A) message might have a MsgSeqNum(34) value that was higher than the next expected MsgSeqNum(34) value, and that in this case, the application should immediately follow the Logon(A) response with a ResendRequest(2) message. Finally, we should move the [`sendResendRequest()`](https://github.com/paritytrading/philadelphia/blob/main/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java#L729) method from `MessageHandler` into the `FIXConnection` class to make it a bit easier for applications to do that. What do you think about this plan?

dmech:
Sorry, I messed things up while syncing my fork to main. This pull request should not have been closed. I will probable have to create a new one.

jvirtanen:
I think you can reopen it just by pushing to the original branch, `main`. 🙂

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
dmech: I am fine with the message in the Logout instead of the connection termination.
Can you confirm, we still want to close the connection after the Logout as stated in the FIX-docs? 

> followed by a termination of the transport layer connection

jvirtanen: Yes, let's have the `statusListener.close(...)` call there as you suggested above. 👍🏼

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
dmech: This sounds like a good plan and it seems to be the only plan if we do not want to handle this logon/resend path in a synchronized way within the library.
I will modify the pull request :thumbsup:

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnectionStatusHandler.java
```diff
@@ -102,14 +102,21 @@ private void handleMsgSeqNum(FIXMessage message, FIXValue msgType, long msgSeqNu
...
         if (msgSeqNum < rxMsgSeqNum)
             handleTooLowMsgSeqNum(message, msgType, msgSeqNum);
-        else
-            sendResendRequest(rxMsgSeqNum);
+        else {
+            if (msgType.contentEquals(Logon))
+                handleLogon(message);
+            else
+                connection.sendResendRequest(rxMsgSeqNum);
+        }
```
jvirtanen, line 110: Let's add braces also around the initial `if` block for balance:
```java
if (msgSeqNum < rxMsgSeqNum) {
    handleTooLowMsgSeqNum(message, msgType, msgSeqNum);
} else {
    ...
```

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnectionStatusListener.java
```diff
@@ -61,6 +61,15 @@ public interface FIXConnectionStatusListener {
...
      * Receive a Logon(A) message.
+     * <p>Please note that the incoming Logon(A) message might have a MsgSeqNum(34)
+     * value that is higher than the next expected MsgSeqNum(34) value.
+     * In this case, the application should immediately follow the Logon(A)
+     * response with a ResendRequest(2) message.
+     * <p>Also the standard suggests:<br>
+     * The initiator and acceptor should wait a short period of time following
+     * receipt of the Logon(35=A) message from the counterparty before
+     * transmitting queued or new application messages to permit both sides
+     * to synchronize the FIX session.
```
jvirtanen, line 72: Let's drop this latter section, as it's a verbatim copy from the specification. If we kept it, we would need to mark it appropriately as a direct quote, but as it is part of the specification, I think we don't need to repeat it here.

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnectionStatusListener.java
```diff
@@ -61,6 +61,15 @@ public interface FIXConnectionStatusListener {
...
      *
      * @param connection the connection
      * @param message the Logon(A) message
```
jvirtanen, line 75: Let's add a reference to the new `FIXConnection#sendResendRequest()` method to this Javadoc after the `@throws` tag:
```java
/**
 * ...
 * @throws IOException if an I/O error occurs
 * @see FIXConnection#resendRequest
 */
```

libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -612,6 +612,15 @@ private void sendTestRequest(FIXValue testReqId) throws IOException {
...
     }
 
+    public void sendResendRequest(long beginSeqNo) throws IOException {
+        prepare(adminMessage, ResendRequest);
+
+        adminMessage.addField(BeginSeqNo).setInt(beginSeqNo);
+        adminMessage.addField(EndSeqNo).setInt(0);
+
+        send(adminMessage);
+    }
```
jvirtanen, line 622: Let's move this after the `sendLogon` method and add an appropriate Javadoc comment. For example:
```java
/**
 * Send a ResendRequest(2) message.
 *
 * @param beginSeqNo the BeginSeqNo(7)
 * @throws IOException if an I/O error occurs
 */
 public void sendResendRequest(long beginSeqNo) throws IOException {
     ...
```

jvirtanen (changes requested):
Nice! 🙌🏼 This is looking good, just see below for a few mostly stylistic nits.

dmech:
Thank you @jvirtanen :raised_hands:
I keep learning. All your 'Nits' are legit points. I leave it to you to resolve the conversations.

jvirtanen (approved):
Thank you, @dmech! 🙌🏼

PR #321, opened 2022-10-16
philadelphia-core: Make 'FIXConnectionStatusListener' optional

jvirtanen:
Make it possible for applications to implement their own administrative message handling.

### DIFF
--- a/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
+++ b/libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
@@ -103,8 +103,33 @@ public class FIXConnection implements Closeable {
      * @see SocketChannel
      */
     public <CHANNEL extends ReadableByteChannel & GatheringByteChannel>
-    FIXConnection(CHANNEL channel, FIXConfig config, FIXMessageListener listener, FIXConnectionStatusListener statusListener, long currentTimeMillis) {
+    FIXConnection(CHANNEL channel, FIXConfig config, FIXMessageListener listener,
+            FIXConnectionStatusListener statusListener, long currentTimeMillis) {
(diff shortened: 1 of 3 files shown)


libraries/core/src/main/java/com/paritytrading/philadelphia/FIXConnection.java
```diff
@@ -135,7 +158,10 @@ public FIXConnection(ReadableByteChannel inboundChannel, GatheringByteChannel ou
         this.senderCompId = config.getSenderCompID();
         this.targetCompId = config.getTargetCompID();
 
-        this.parser = new FIXMessageParser(config, new MessageHandler(listener));
+        if (statusListener != null)
+            listener = new FIXConnectionHandler(config, this, listener, statusListener);
```
ddimtirov, line 162: I assume there will be subsequent PR getting rid of this?

It is reasonable to have the client specify what level of session handling they expect - if you rename it to `BasicFixSessionHandler` and make all private methods protected, it will be overall improvement in usability.

jvirtanen: Actually my thought was to leave it like this. Either an application opts into the default administrative message handling by passing in a `FIXConnectionStatusListener` or it handles all inbound messages by itself.

The default administrative message handling, now in `FIXConnectionStatusHandler`, is not a great base class if one would like to support e.g. actual message recovery. We also want to retain the option to change its implementation as needed, hence preferring to keep it as package-private. Right now you could describe the class as follows: "Optimize for low latency, and do the least you can get away with."

I think what you're suggesting makes sense, but I would do it later (i.e. after 2.0.0) via new interfaces rather than the current `FIXConnectionStatusHandler` and `FIXConnectionStatusListener`. What do you think?

PR #325, opened 2022-10-23
philadelphia-core: Clean up terminology

jvirtanen:
Use `inMsgSeqNum` and `outMsgSeqNum` for consistency.


jvirtanen:
This is pending on #320.

PR #361, opened 2023-03-04
philadelphia-generate: Clean up type hints

jvirtanen:


jvirtanen:
Let's get back to this October 2024 when Python 3.8 reaches end-of-life.

PR #503, opened 2024-09-05

yrousset:


JanStureNielsen (commented):
Hi @yrousset -- your changes include lots of Eclipse configuration files - is that correct? This project prefers to keep developer tool artifacts out of the repository so those changes should be removed from your PR which will also make it much easier to see the scope of your proposal. (FWIW, I use a global ignore file to avoid adding tooling-specific files to repos.)

PR #680, opened 2026-07-21
core: Improve 'FIXValue#asInt()'

jvirtanen:
Throw `FIXValueFormatException` if
* the value consists of zero bytes,
* the value consists of a single `-` byte, or
* the value does not fit into a `long`.


jvirtanen:
This is based on [Tagline](https://github.com/jvirtanen/tagline)'s Int data type implementation.

